PR #12, opened 2024-03-19
Feature/firebase2

antomax33:
The basic register screen from the bootcamp to go to an almost empty second screen.

I added route to make it cleaner and a navigate function with basic navigate function to keep track of navigation and to add test for the future.


RandomUsername1315 (approved):
LGTM

PR #24, opened 2024-03-22
Add implementation tests to CI

RandomUsername1315:


.github/workflows/ci.yml
```diff
@@ -138,44 +144,83 @@ jobs:
...
+          echo "compile-tests,${{ steps.compile-tests.outcome }}" >> grading-result.csv
+          echo "instrumentation-tests-task,${{ steps.instrumentation-tests.outcome }}" >> grading-result.csv
+          echo "coverage-task,${{ steps.coverage.outcome }}" >> grading-result.csv
+          
+          # Tests results
+          
+          result_file="app/build/outputs/androidTest-results/connected/*/TEST-*.xml"
+          
+          test_names=("useAppContext")
+          test_classes=("com.github.se.bootcamp.todo.ExampleInstrumentedTest")
```
antomax33, line 165: Do you only test ExampleInstrumentedTest or start from this one?

antomax33 (commented):
If it start the tests and get the coverage, it is good for me

antomax33 (approved):

.github/workflows/ci.yml
RandomUsername1315: I used this to explicitly check if this test is run, to fetch the results. But normally all the tests in androidTest should be run

PR #39, opened 2024-03-27
Add Meal and Ingredient Data Classes

RandomUsername1315:
# Pull Request Overview

This pull request arises from the strategy of branching out to facilitate the code review process.

## Changes

- **Added `Meal` Data Class**: This class models a meal within our application, encapsulating multiple ingredients.
- **Added `Ingredient` Data Class**: Introduces an ingredient as a component of a meal, with nutritional details such as calories and fat content.
- **Added Tests for `Meal` and `Ingredient`**: These tests validate the functionality of the `Meal` and `Ingredient` classes as expected.
- **Added Necessary Subclasses**: These subclasses are introduced to reduce mental load, enhancing code readability and maintainability. (see #51 )

These enhancements are pivotal for the forthcoming Firebase database integration.

## Implementation Checklist

- [x] Ingredient
- [x] Meal

## Tests Checklist

- [x] Ingredient
- [x] Meal


app/src/main/java/com/github/se/polyfit/model/meal/Meal.kt
```diff
@@ -0,0 +1,88 @@
...
+     *
+     * @param data The Meal object to serialize.
+     * @return A pair of maps. The first map contains the Meal properties, and the second a list of
+     *   map that contains the Ingredient properties.
+     */
+    fun serializeMeal(data: Meal): Pair<Map<String, Any>, List<Map<String, Any>>> {
+      val map = mutableMapOf<String, Any>()
+
+      // Convert each property manually
+      map["uid"] = data.uid
```
antomax33, line 55: Same hardcode value

antomax33 (approved):
It looks good,

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @property name is unique for each meal
+ */
+class Ingredient(
+    val name: String, // name is now a val
+    servingSize: Double,
+    calories: Double,
+    protein: Double,
+    carbohydrates: Double,
+    fat: Double
```
medwardson, line 18: Ingredients will likely have a lot more nutritional information (see this [design](https://www.figma.com/file/JAs7WfDzS8lFij1xYd2J1I/User-Stories?type=design&node-id=60-515&mode=design&t=2aFeg5xv7Ml8Dtfl-4) for example. I'm wondering thoughts on creating a `NutritionalInformation` class, or something similar? Also, we might have different size formats (e.g. ml, g, 1 unit (i.e. for an oreo for example), so it may be worth creating an enum of sizeTypes, and maybe also a map for converting between similar ones (e.g. servingSize = 0.2 and sizeType = g, so we know this is 200mg). Happy to discuss more just initial thoughts

app/src/main/java/com/github/se/polyfit/model/meal/Meal.kt
```diff
@@ -0,0 +1,88 @@
+package com.github.se.polyfit.model.meal
+
+import com.github.se.polyfit.model.ingredient.Ingredient
+
+// modeled after the log meal api
+data class Meal(
+    val occasion: MealOccasion,
```
medwardson, line 7: maybe I'm looking ahead too far, but we can potentially add an optional temperature field to Meal for in the future suggesting food based on similar temperatures

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.github.se.polyfit.model.meal.Meal
+import kotlin.properties.Delegates
+
+/**
+ * Represents an ingredient in a meal. Automatically updates the parent meal when any property
+ * changes.
+ *
+ * @property name is unique for each meal
+ */
+class Ingredient(
```
medwardson, line 12: we might want to abstract this into two classes

an ingredient is something that is "normalized" into base unit, so if its grams it will be 1g, or something similar. Then, we have a second class called serving, each serving will have an ingredient and a serving size for example. This way we can reuse ingredients rather than create a new object each time for the same food. it'll save space because the ingredient object has a lot of fields, but a serving object doesn't need as much

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
```diff
@@ -0,0 +1,74 @@
...
+) {
+  var servingSize: Double by
+      Delegates.observable(servingSize) { _, _, _ -> parentMeal?.updateMeal() }
+  var calories: Double by Delegates.observable(calories) { _, _, _ -> parentMeal?.updateMeal() }
+  var protein: Double by Delegates.observable(protein) { _, _, _ -> parentMeal?.updateMeal() }
+  var carbohydrates: Double by
+      Delegates.observable(carbohydrates) { _, _, _ -> parentMeal?.updateMeal() }
+  var fat: Double by Delegates.observable(fat) { _, _, _ -> parentMeal?.updateMeal() }
+
+  var parentMeal: Meal? = null
```
medwardson, line 28: Then ingredient will belong to serving instead of a meal, and it can belong to many servints not just one

medwardson (commented):
Good first pass, left some comments about structuring the data a bit differently, lmk what your thoughts on this are, happy to discuss more

medwardson (changes requested):
just marking as request changes so we can discuss structure before merging

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Represents an ingredient in a meal. Automatically updates the parent meal when any property
+ * changes.
+ *
+ * @property name is unique for each meal
+ */
+class Ingredient(
+    val name: String, // name is now a val
+    servingSize: Double,
+    calories: Double,
```
aprilslis, line 15: to be discussed: we might want to store more than just the basics (calories, protein, carbs, etc), for example sub section under fat

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
RandomUsername1315: Exactly what I was thinking, I tried an implementation along those lines. But in order to implement something like that, the auto update feature of the meal when an new ingredient is added (at least the current implementation).

Did you already have an interface/abstract class design in mind ?

RandomUsername1315: Additionally, the issue with the API I have in mind (LogMeal) is that the documentation lacks detail. It does not describe which nutritional values are returned. This is why it is modeled in a simplistic way. I am waiting for @Sobell13 to provide more information about that.

app/src/main/java/com/github/se/polyfit/model/meal/Meal.kt
RandomUsername1315: I attempted to model the classes based on the API results, but I could definitely include that. Were you thinking more along the lines of a range or an enum for temperature, like cold, medium, hot?

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
RandomUsername1315: Currently, it is modeled based on the information provided in the documentation about the nutritional values returned. At this moment, I am not entirely sure which API to use (and what information it can return), which is why it models only simple nutritional facts for now. But I can be expanded in the future.

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
RandomUsername1315: Here is a sample response from the most likely API we will use: 
{
    "foodName": [
        "meat lasagna"
    ],
    "hasNutritionalInfo": true,
    "ids": [
        1034
    ],
    "imageId": 1368311,
    "nutritional_info": {
        "calories": 602.5,
        "totalNutrients": {
            "CA": {
                "label": "Calcium",
                "quantity": 527.01,
                "unit": "mg"
            },
            "CHOCDF": {
                "label": "Carbs",
                "quantity": 34.72,
                "unit": "g"
            },
            "CHOLE": {
                "label": "Cholesterol",
                "quantity": 165.85,
                "unit": "mg"
            },
            "ENERC_KCAL": {
                "label": "Energy",
                "quantity": 602.5,
                "unit": "kcal"
            },
            "FAMS": {
                "label": "Monounsaturated fats",
                "quantity": 10.67,
                "unit": "g"
            },
            "FAPU": {
                "label": "Polyunsaturated",
                "quantity": 2.57,
                "unit": "g"
            },
            "FASAT": {
                "label": "Saturated",
                "quantity": 15.29,
                "unit": "g"
            },
            "FAT": {
                "label": "Fat",
                "quantity": 32.45,
                "unit": "g"
            },
            "FATRN": {
                "label": "Trans fat",
                "quantity": 0.31,
                "unit": "g"
            },
            "FE": {
                "label": "Iron",
                "quantity": 4.6,
                "unit": "mg"
            },
            "FIBTG": {
                "label": "Fiber",
                "quantity": 3.92,
                "unit": "g"
            },
            "FOLAC": {
                "label": "Folic acid",
                "quantity": 34.24,
                "unit": "µg"
            },
            "FOLDFE": {
                "label": "Folate equivalent (total)",
                "quantity": 100.81,
                "unit": "µg"
            },
            "FOLFD": {
                "label": "Folate (food)",
                "quantity": 42.7,
                "unit": "µg"
            },
            "K": {
                "label": "Potassium",
                "quantity": 1041.44,
                "unit": "mg"
            },
            "MG": {
                "label": "Magnesium",
                "quantity": 75.08,
                "unit": "mg"
            },
            "NA": {
                "label": "Sodium",
                "quantity": 1575.92,
                "unit": "mg"
            },
            "NIA": {
                "label": "Niacin (B3)",
                "quantity": 6.26,
                "unit": "mg"
            },
            "P": {
                "label": "Phosphorus",
                "quantity": 609.72,
                "unit": "mg"
            },
            "PROCNT": {
                "label": "Protein",
                "quantity": 43.91,
                "unit": "g"
            },
            "RIBF": {
                "label": "Riboflavin (B2)",
                "quantity": 0.69,
                "unit": "mg"
            },
            "SUGAR": {
                "label": "Sugars",
                "quantity": 11.27,
                "unit": "g"
            },
            "SUGAR.added": {
                "label": "Sugars, added",
                "quantity": 0.0,
                "unit": "g"
            },
            "THIA": {
                "label": "Thiamin (B1)",
                "quantity": 0.3,
                "unit": "mg"
            },
            "TOCPHA": {
                "label": "Vitamin E",
                "quantity": 3.48,
                "unit": "mg"
            },
            "VITA_RAE": {
                "label": "Vitamin A",
                "quantity": 246.13,
                "unit": "µg"
            },
            "VITB12": {
                "label": "Vitamin B12",
                "quantity": 2.68,
                "unit": "µg"
            },
            "VITB6A": {
                "label": "Vitamin B6",
                "quantity": 0.47,
                "unit": "mg"
            },
            "VITC": {
                "label": "Vitamin C",
                "quantity": 14.78,
                "unit": "mg"
            },
            "VITD": {
                "label": "Vitamin D",
                "quantity": 40.8,
                "unit": "µg"
            },
            "VITK1": {
                "label": "Vitamin K",
                "quantity": 16.54,
                "unit": "µg"
            },
            "ZN": {
                "label": "Zinc",
                "quantity": 5.82,
                "unit": "mg"
            }
        }
    },
    "serving_size": 385.2
}

medwardson: Here's an idea in pseudocode but not 100% sure:

```kt
enum class MeasurementUnit {
    Kilogram, // could also do gram and milliliter, but i think best to keep this same for math simplicity later
    Litre,
    Unit // For standard sized items, i.e. 1 oreo, 1 lollipop
}

class Ingredient (
  val name: String,
  val unitType: MeasurementUnit,
  // All of these values should be normalized to the unit type. So if we know something is 10 calories for 100ml, then this should be 100 calories since it is liters.
  val calories: Double,
  val carbohydrates: Double,
  // Include all nutritional info we want to store here
) { ... }

class Serving (
  val ingredient: Ingredient,
  val servingSize: Double // This will be in terms of base units. So, if someone has 100ml then its 0.1 for litres
  val meal: Meal
) { ... }

class Meal (
  val occasion: MealOccasion,
  val name: String,
  private val servings: MutableList<Serving> = mutableListOf()
) { ... }
```

and then we can maybe have something calculate the totals for meal?

medwardson: I think we can set it up for now with ingredients just having basic macros and add more after we know more about the API from alex

app/src/main/java/com/github/se/polyfit/model/meal/Meal.kt
medwardson: i had in mind an integer representing temp in celsius so we can have more precision control, but open to either

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
medwardson: The nice thing about structuring the ingredients this way is we can always check our DB for the ingredient before we send the API request out, so that eventually we will build our own database of nutritional info

RandomUsername1315: In order to get the ingredients that is another API call , but the unit for the response is just weight (haha). So I am guessing grames ??
Here is the structured data for a dish, specifically "meat lasagna":

```json
{
    "dish_id": [1034],
    "foodName": ["meat lasagna"],
    "imageId": 1368311,
    "recipe": [
        {
            "id": 522,
            "name": "beef",
            "weight": 30.74
        },
        {
            "id": 807,
            "name": "pork sausage",
            "weight": 22.94
        },
        {
            "id": 750,
            "name": "tomato sauce",
            "weight": 74.34
        },
        {
            "id": 112,
            "name": "tomato",
            "weight": 39.75
        },
        {
            "id": 381,
            "name": "garlic",
            "weight": 0.61
        },
        {
            "id": 47,
            "name": "sugar",
            "weight": 0.84
        },
        {
            "id": 925,
            "name": "italian seasoning",
            "weight": 0.18
        },
        {
            "id": 45,
            "name": "salt",
            "weight": 0.61
        },
        {
            "id": 175,
            "name": "pepper",
            "weight": 0.12
        },
        {
            "id": 569,
            "name": "egg",
            "weight": 15.17
        },
        {
            "id": 256,
            "name": "parsley",
            "weight": 0.49
        },
        {
            "id": 862,
            "name": "cottage cheese",
            "weight": 65.99
        },
        {
            "id": 837,
            "name": "ricotta cheese",
            "weight": 24.88
        },
        {
            "id": 18,
            "name": "parmesan cheese",
            "weight": 5.06
        },
        {
            "id": 926,
            "name": "lasagna noodles",
            "weight": 51.88
        },
        {
            "id": 927,
            "name": "provolone cheese",
            "weight": 17.2
        },
        {
            "id": 97,
            "name": "mozzarella",
            "weight": 34.41
        }
    ],
    "serving_size": 385.2
}

medwardson: can you link me the docs? i agree this is a weird response body lol

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
RandomUsername1315: The documentation sucks. You can find the list of nutriments for each meal https://github.com/swent-group10/polyfit/pull/39#discussion_r1541962229 but would you rather store all of them or a particular subset?

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
RandomUsername1315: So to resume this huge spam: For meals, we get loads of nutrition facts on nutrients and macros, but with all kinds of different measurements. As for ingredients, the API just gives us their weight and that's about it—no clue on what's actually inside them.

But hey, there's another cool API over at https://spoonacular.com/food-api/docs#Get-Recipe-Information that can break down the ingredients if you throw a recipe at it.

RandomUsername1315: Yeah, this API is much better. It gives you a breakdown of the recipe's nutritional info and other details, including the nutritional information about the ingredients (https://spoonacular.com/food-api/docs#Nutrition-by-ID).

It has an API for recipes and more. The only downside is that its image recognition is limited. But I think that's okay for the scope of this project. What do you think, @medwardson?

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
RandomUsername1315: Here is actually the API we're going to use. What nutritional values do you think would be a good idea to be stored?
https://spoonacular.com/food-api/docs#Nutrition-by-ID
@aprilslis

medwardson (approved):

aprilslis (approved):

PR #42, opened 2024-03-27
Feature/UI/register screen

antomax33:
I added the UI of the register screen.


app/src/main/java/com/github/se/polyfit/ui/compose/Compose.kt
medwardson, line 1: i think if this file is basically just being used for title it should be named as `Title.kt` for clarity, as `Compose.kt` is unclear what this is

medwardson (approved):
lgtm, small suggestions added

build.gradle.kts
```diff
@@ -5,7 +5,7 @@ buildscript {
 }
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 plugins {
-    id("com.android.application") version "8.3.1" apply false
+    id("com.android.application") version "8.2.2" apply false
```
aprilslis, line 8: if version 8.2.2 does not work for you, you might want to just update your Android Studio IDE

app/src/main/java/com/github/se/polyfit/ui/theme/Color.kt
```diff
@@ -9,3 +9,5 @@ val Pink80 = Color(0xFFEFB8C8)
 val Purple40 = Color(0xFF6650a4)
 val PurpleGrey40 = Color(0xFF625b71)
 val Pink40 = Color(0xFF7D5260)
+val BlueButton = Color(0xFF7158E2) // I quickly to match the figma, I don't know which material it is.
+val PinkTitle = Color(0xFFCD84F1)
```
aprilslis: Might want to name these better. The blue/purple and pink are both in buttons, bottom navigation bars and other parts. Maybe something like primaryPurple?

app/src/main/java/com/github/se/polyfit/ui/screen/LoginScreen.kt
```diff
@@ -58,43 +66,60 @@ fun LoginScreen(navController: Navigation) {
       modifier = Modifier.fillMaxSize().testTag("LoginScreen"),
       color = MaterialTheme.colorScheme.background) {
         Column(
-            modifier = Modifier.padding(24.dp),
             verticalArrangement = Arrangement.Center,
             horizontalAlignment = Alignment.CenterHorizontally) {
-              Spacer(Modifier.weight(0.15f))
+              Spacer(Modifier.weight(0.2f))
+
+              Titre(Modifier.testTag("LoginTitle"), 50.sp)
```
aprilslis, line 73: titre?

app/src/main/java/com/github/se/polyfit/ui/theme/Color.kt
```diff
@@ -9,3 +9,6 @@ val Pink80 = Color(0xFFEFB8C8)
 val Purple40 = Color(0xFF6650a4)
 val PurpleGrey40 = Color(0xFF625b71)
 val Pink40 = Color(0xFF7D5260)
+val BlueButton =
+    Color(0xFF7158E2) // I quickly to match the figma, I don't know which material it is.
```
aprilslis, line 13: nit: is there any reason why it is not on the same line

app/src/main/java/com/github/se/polyfit/ui/compose/Compose.kt
antomax33: I think it will be used for more than the title, like all Composable used in different screen and the gradiant used everywhere

aprilslis (commented):
lgtm, small nits addressed in comments

app/src/main/java/com/github/se/polyfit/ui/screen/LoginScreen.kt
antomax33: Yeah my bad, it's the french word for title

app/src/main/java/com/github/se/polyfit/ui/theme/Color.kt
antomax33: Nop, the comment too long and ktfmt

app/src/main/java/com/github/se/polyfit/ui/screen/LoginScreen.kt
```diff
@@ -58,43 +66,60 @@ fun LoginScreen(navController: Navigation) {
...
+        Image(
+            painter = painterResource(id = R.drawable.google_logo),
+            contentDescription = "Google_logo",
+            contentScale = ContentScale.Fit,
+            modifier = imageModifierGoogle)
+        Text(text = "Sign in with Google")
+      }
+}
+
 class Authentication(private val navigation: Navigation, private val context: Context) {
```
RandomUsername1315, line 123: If we try to follow the view, viewmodel, model pattern, the Authentification class should belong its own class in the model folder. This  would improve the readability of the code

app/src/main/java/com/github/se/polyfit/ui/screen/LoginScreen.kt
```diff
@@ -126,3 +151,10 @@ class Authentication(private val navigation: Navigation, private val context: Co
     }
```
RandomUsername1315, line 151: I just noticed that response might be null here:
```kotlin
val response = result.idpResponse
Log.i("LoginScreen", "response: $response")
if (result.resultCode == RESULT_OK) {
``` 
that it might be a good idea to if response is null to still log it as an error (line 149)

RandomUsername1315: It also might be a good idea to be consistent in the type of login used. Here you used information (login.i) and above debug. For similar cases

build.gradle.kts
RandomUsername1315: @aprilslis what is wrong with version 8.3.1 ?

PR #56, opened 2024-04-07
Setting up api calls Spoonacular 

RandomUsername1315:
# Pull Request Overview

This pull request will implement the correct API interface to request information from Spoonacular.

## Changes

- **Added `SpoonacularApiCaller` Object**: This class contains all the necessary functions to request data.
- **Added `ImageClassificationResponse`, `ImageAnalysisResponse` Data Classes**: These represent the potential responses from the API.
- **Added Tests for `SpoonacularApiCaller`**.

## Implementation Checklist

- [x] ImageAnalysisResponse
- [x] ImageClassificationResponse
- [x] RecipeNutrition
- [x] SpoonacularApiCaller

## Tests Checklist

- [x] SpoonacularApiCaller


RandomUsername1315:
Issue with CI. Does not run.

RandomUsername1315:
Dont worry, the 4,600 new lines is due to a json used as an example that is 4,000 lines.

app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt
```diff
@@ -0,0 +1,106 @@
...
+
+      val responseBody = response.body?.string() ?: ""
+      val jsonObject = JSONObject(responseBody)
+      RecipeNutritionResponseAPI.fromJsonObject(jsonObject)
+    } catch (e: IOException) {
+      Log.e("SpoonacularApiCaller", "Error getting recipe nutrition", e)
+      throw Exception("Error getting recipe nutrition : ", e)
+    }
+  }
+}
```
Neb1010, line 106: seems good, I don't find the API key so ensure there's no API key in vc

app/src/main/java/com/github/se/polyfit/model/ingredient/Ingredient.kt
```diff
@@ -31,11 +36,12 @@ data class Ingredient(
       return try {
```
Neb1010, line 36: maybe add as? to return null if the value cannot be cast

app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCaller.kt
```diff
@@ -0,0 +1,4033 @@
+import android.util.Log
```
Neb1010, line 1: I would say in general in the test make sure for deserialization if the data we receive are not conform or what expected like null or "corrupt" are well tested for robustess

Neb1010 (commented):
First review, if you write some improvement tag me for a second review.

app/src/main/java/com/github/se/polyfit/data/api/RecipeNutritionResponseAPI.kt
```diff
@@ -0,0 +1,64 @@
...
+
+      val ingredients = mutableListOf<Ingredient>()
+
+      for (i in 0 until ingredientsJsonArray.length()) {
+        val ingredientJsonObject = ingredientsJsonArray.getJSONObject(i)
+        val nutritionalInformation = NutritionalInformation(mutableListOf())
+
+        val nutrientsJSONArray = ingredientJsonObject.getJSONArray("nutrients")
+
+        for (j in 0 until nutrientsJSONArray.length()) {
```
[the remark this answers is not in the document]
RandomUsername1315: Nice suggestion

app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt
RandomUsername1315: Yeah that is normal. The api key is stored in the local.properties which is not push to the git. I will send the keys  in the group chat when this implementation is accepted to that everyone can copy them locally.

RandomUsername1315 (commented):
Thanks for the feedback.

RandomUsername1315:
@Neb1010, thank you for your review. When you do a second pass, could you give me your opinion on the way that errors are implemented? Using the status as a way to indicate whether the transaction succeeded or not allows the MealViewModel to handle different cases, decoupling the API (this class) from other logic.

app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt
```diff
@@ -0,0 +1,106 @@
...
+    val requestBody = file.asRequestBody(mediaType)
+    val body =
+        MultipartBody.Builder()
+            .setType(MultipartBody.FORM)
+            .addFormDataPart("file", file.name, requestBody)
+            .build()
+
+    val request =
+        Request.Builder()
+            .url(API_URL + IMAGE_ANALYSIS_ENDPOINT)
```
sharif-bbd, line 47: This works for the concatenation but could raise issues if there are trailing slashes on the baseUrl or leading slashes on the endpoint. Althoug it is fine if this baseUrl and these endpoints are the only ones that will be used. But you could make this problem go away by using a URL Builder
```
val httpUrl = HttpUrl.Builder()
    .scheme("https") 
    .host("spoonacular-recipe-food-nutrition-v1.p.rapidapi.com") 
    .addPathSegment(endpoint) 
    .build()

**Review by sharif-bbd — APPROVED** [review]

**Review by Neb1010 — APPROVED** [review]

File: app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt [hunk 6]
```diff
@@ -0,0 +1,106 @@
...
+            .post(body)
+            .addHeader("content-type", "multipart/form-data; boundary=---011000010111000001101001")
```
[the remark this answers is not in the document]
RandomUsername1315: I tried to follow the recommended approach as cited in the documentation ([Spoonacular API](https://rapidapi.com/spoonacular/api/recipe-food-nutrition)). I tested it directly with the API and it works. However, I agree with you that it is a strange approach.



```kotlin
val client = OkHttpClient()

val mediaType = MediaType.parse("multipart/form-data; boundary=---011000010111000001101001")
val body = RequestBody.create(mediaType, "-----011000010111000001101001\r\nContent-Disposition: form-data; name=\"file\"\r\n\r\n\r\n-----011000010111000001101001--\r\n\r\n")
val request = Request.Builder()
	.url("https://spoonacular-recipe-food-nutrition-v1.p.rapidapi.com/food/images/classify")
	.post(body)
	.addHeader("content-type", "multipart/form-data; boundary=---011000010111000001101001")
	.addHeader("X-RapidAPI-Key", "")
	.addHeader("X-RapidAPI-Host", "spoonacular-recipe-food-nutrition-v1.p.rapidapi.com")
	.build()

val response = client.newCall(request).execute()
```
But it seems that you're right. I tested it with live API and it works without needing to specify the content type.

app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt
RandomUsername1315: Thank you for the suggestion; I'll take it into consideration. Incorporating this into the class could be challenging, as it needs to support both HTTP and HTTPS connections. This might significantly increase the class's complexity.

PR #72, opened 2024-04-09
Feature/top app bar

Neb1010:
-An implementation of a Top Bar that can have a navigation drawer and actions Icons.
-Icons that can be used 
-data user google account for having account info


app/src/main/java/com/github/se/polyfit/ui/utils/UiExtrasDev.kt
```diff
@@ -0,0 +1,19 @@
+package com.github.se.polyfit.ui.utils
+
+import androidx.compose.material3.AlertDialog
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.runtime.Composable
+
+@Composable
+fun FunctionalityNotAvailablePopup(onDismiss: () -> Unit) {
```
medwardson, line 10: nice 👍🏼

app/src/main/java/com/github/se/polyfit/ui/components/TopAppBar.kt
medwardson, line 1: These look good, but I don't understand what the use for them is as they don't line up with our designs in Figma - Where is this supposed to be used?

Neb1010: Yeah sorry, I've checked everywhere telegram, discord even slacked (lol) I haven't found where the design decision are taken, so please give me the link where design decisions are taken, my bad I probably didn't pay attention about where it was discussed. But I've added on figma quickly what it would looks like (still working on it). Because recent app use that and I talked with Charles and he said it was a realy good idea. and what

app/src/androidTest/java/com/github/se/polyfit/ui/components/GradientBoxScreen.kt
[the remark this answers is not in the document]
Neb1010: Yeah, In fact as probably nobody red how to apply color correctly and I was the alone doing in M3 way, when I merge with other branch for continuity, I got error with people hard coding color so I had to make some changement on my side and just probably a mistake of pushing it into the CI, even I paid attention do not push what isn't mine, I don't know

app/src/main/java/com/github/se/polyfit/ui/components/IconsEx.kt
```diff
@@ -0,0 +1,105 @@
...
+  // to replace with app logo or whatever
+  Icon(
+      imageVector = Icons.AutoMirrored.Outlined.MenuOpen,
+      modifier = modifier.clickable(onClick = onClick),
+      tint = MaterialTheme.colorScheme.onSurfaceVariant,
+      contentDescription = "App Icon")
+}
+
+@Composable
+fun SearchIcon(onClick: () -> Unit) {
```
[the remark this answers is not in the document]
Neb1010: This is just an implementation for if we need it for future

PR #74, opened 2024-04-09
Fixes issues with checking if login intent was sent

RandomUsername1315:
## Summary

This pull request addresses a failing test case in the `LoginTest` class, specifically the `googleSignInReturnsValidActivityResult` test. The test was failing due to an `IllegalStateException` indicating that no compose hierarchies were found in the app.

### Issue and Root Cause

The test failed with an `IllegalStateException`, suggesting issues with `setContent` related operations. It was determined that conditions were not properly initialized, and Compose UI elements were not stabilized before the test ran.

### Implemented Solution

The issue has been resolved by implementing a waiting mechanism that ensures all elements are stabilized before the test is executed.


Neb1010 (approved):
Yeah leeeeeeeeeet's go

PR #75, opened 2024-04-09
authentication test failing

RandomUsername1315:
## Summary

This pull request addresses a failing test case in the `LoginTest` class, specifically the `googleSignInReturnsValidActivityResult` test. The test was failing due to an `IllegalStateException` indicating that no compose hierarchies were found in the app.

## Issue and Root Cause

The test failed with an `IllegalStateException`, suggesting issues with `setContent` related operations. It was determined that conditions were not properly initialized, and Compose UI elements were not stabilized before the test ran.

## Implemented Solution

The issue has been resolved by implementing a waiting mechanism that ensures all elements are stabilized before the test is executed.

## Update

It turns out the issue was due to incorrect initialization of the intents in the test. To ensure more consistent testing, the intents are now mocked.

### DIFF
--- a/app/src/androidTest/java/com/github/se/polyfit/ui/screen/LoginTest.kt
+++ b/app/src/androidTest/java/com/github/se/polyfit/ui/screen/LoginTest.kt
@@ -70,7 +70,6 @@ class LoginTest : TestCase() {
   @Test
   fun googleSignInReturnsValidActivityResult() {
 
-
     ComposeScreen.onComposeScreen<LoginScreen>(composeTestRule) {
       loginButton {
         assertIsDisplayed()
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/com/github/se/polyfit/ui/screen/LoginTest.kt
```diff
@@ -17,7 +19,15 @@ import org.junit.runner.RunWith
...
   @get:Rule val composeTestRule = createAndroidComposeRule<MainActivity>()
 
-  @get:Rule val intentsTestRule = IntentsTestRule(MainActivity::class.java)
+  @Before
+  fun setup() {
+    Intents.init()
+  }
+
+  @After
+  fun tearDown() {
```
Neb1010, line 28: many **_tears_** run **_down_** my cheek

Neb1010 (approved):

PR #79, opened 2024-04-10
Backend/user class

Neb1010:
User class
You can have
- account id
-  name 
- familyName 
- - both in displayName 
- mail adress
- google account image


app/src/main/java/com/github/se/polyfit/data/user/User.kt
```diff
@@ -0,0 +1,20 @@
...
+data class User(
+    val id: String?,
+    val displayName: String?,
+    val familyName: String?,
+    val getGivenName: String?,
+    val email: String?,
+    val photoURL: Uri?
+) {
+  companion object {
+    fun resetCurrentUser() {
```
RandomUsername1315, line 14: Instead of setting the pointer to null, it could be really useful to have a method that resets the fields to default values. Did you add this method in anticipation of implementing a logout feature?

app/src/main/java/com/github/se/polyfit/ui/utils/Authentication.kt
```diff
@@ -44,6 +45,16 @@ class AuthenticationCloud @Inject constructor(private val context: Context) : Au
     Log.i("LoginScreen", "response: $response")
     if (result.resultCode == Activity.RESULT_OK) {
       Log.i("LoginScreen", "User signed in")
+      // to get google acount infos
+      val account = GoogleSignIn.getLastSignedInAccount(context)
+      User.currentUser =
```
RandomUsername1315, line 50: - We need to figure out a way to pass this data to other screens, as I'm encountering the same issue in PR #58. Let's get together and discuss a possible implementation.

RandomUsername1315 (approved):
LGTM

PR #92, opened 2024-04-12
Bugfix: Nutritional value and isComplete

aprilslis:
Fixing the double addition of nutritional value and the state for isComplete


medwardson (approved):
nice job keeping PR small

PR #93, opened 2024-04-12
Camera to spoonacular API call

RandomUsername1315:
## Pull Request Description

This pull request integrates the functionality introduced in PR #89 by implementing the logic required to pass data between screens effectively. Specifically, it facilitates the transfer of information from one user interface component to another, ensuring seamless data flow within the application.

## Potential Questions for Review

I am aware that the current method for passing the MealViewModel between screens might not be the most efficient or ideal approach. Could you suggest any alternative strategies or improvements that might enhance the robustness and maintainability of this data transfer process?


antomax33 (approved):
LGTM

app/src/androidTest/java/com/github/se/polyfit/ui/screen/IngredientTest.kt
```diff
@@ -74,6 +73,7 @@ class IngredientTest : TestCase() {
             nutritionalInformation = NutritionalInformation(mutableListOf()),
         )
     every { mockMealViewModel.meal.value } returns testMeal
+    every { mockMealViewModel.meal.value } returns testMeal
```
aprilslis, line 76:
```suggestion
    every { mockMealViewModel.meal.value } returns testMeal
```

app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
```diff
@@ -10,7 +10,7 @@ import com.github.se.polyfit.model.meal.Meal
 class MealViewModel(
     private val userId: String,
     firebaseID: String = "",
-    initialMeal: Meal? = null,
+    var initialMeal: Meal? = null,
```
aprilslis, line 13: Can you explain why we are using it as a var?

app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
```diff
@@ -25,10 +25,17 @@ class MealViewModel(
...
+      _meal.value = initialMeal!!.copy()
     } else {
       _meal.value = Meal.default()
     }
+
+    _meal.observeForever { initialMeal = it }
+  }
+
+  fun setMealData(meal: Meal) {
+    // add random line of code
```
aprilslis, line 37: I don't think this comment makes sense?

app/src/main/java/com/github/se/polyfit/ui/components/IngredientList.kt
```diff
@@ -45,18 +45,24 @@ fun IngredientList(
     it: PaddingValues,
     mealViewModel: MealViewModel,
 ) {
-  val meal by mealViewModel.meal.observeAsState()
-  val ingredients = meal?.ingredients ?: emptyList()
+  //    var ingredients =
+
```
aprilslis, line 49:
```suggestion
```

app/src/main/java/com/github/se/polyfit/ui/components/IngredientList.kt
```diff
@@ -45,18 +45,24 @@ fun IngredientList(
...
   val potentialIngredients = listOf<Ingredient>()
+
   val onAddIngredient = { index: Int -> Log.v("Add Ingredient", "Clicked") }
   val initialSuggestions = 3
   val potentialIndex = remember {
     // equivalent to min(3, potentialIngredients.size)
     mutableIntStateOf(initialSuggestions.coerceAtMost(potentialIngredients.size))
   }
 
+  //    val
```
aprilslis, line 62:
```suggestion
```

app/src/main/java/com/github/se/polyfit/MainActivity.kt
```diff
@@ -34,20 +35,23 @@ class MainActivity : ComponentActivity() {
     controller.systemBarsBehavior =
         WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
 
+    // TO DO: technical debt, next deadline find better way to pass arguments from overview screen
+    // to
+    // add meal screen
```
aprilslis, line 40:
```suggestion
    // to add meal screen
```

app/src/test/java/com/github/se/polyfit/viewmodel/meal/MealViewModelTest.kt
```diff
@@ -74,6 +74,25 @@ class MealViewModelTest {
...
+            mutableListOf(),
+            "testID")
+    mealViewModel = MealViewModel(userID, mealRepo = mealFirebaseRepository)
+
+    mealViewModel.setMealData(meal)
+
+    val updatedMeal = mealViewModel.meal.value!!
+    assertEquals(meal, updatedMeal)
+  }
+
```
aprilslis, line 95: Oops sorry, i changed the meal viewmodel test and the line coverage can be recognized in jacoco now. So you would need to rebase and add this test into the new format

aprilslis (commented):
Looks good, small nits and I think you need to either drop the meal viewmodel test or change the test, but that is all!

app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
RandomUsername1315, line 37:
```suggestion
```

app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
RandomUsername1315: I was waiting for you to commit your changes and did not want to break any implementation before that.

app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt
```diff
@@ -126,7 +125,7 @@ class SpoonacularApiCaller {
     val meal = MutableLiveData<Meal>()
     meal.postValue(Meal.default())
 
-    scope.launch {
+    GlobalScope.launch {
```
sharif-bbd, line 128: This could lead to some issues later on as calling GlobalScope.launch makes it so that the coroutine will never shut down. Although it works from now since this coroutine doen't have any sense of the lifecyle of the activity or feature it is implemented in it will stay running in the background and could cause some issues. 


some alternative could be linking it to a viewModel and call the viewmodelScope.launch coroutine or even lifecycleScope.launch
Another idea would be to create you're own scope this would allow more control if needed.

not needed for now but put a TODO on it to modify it later

app/src/main/java/com/github/se/polyfit/ui/navigation/NavigationGraph.kt
```diff
@@ -7,13 +7,19 @@ import androidx.navigation.compose.composable
...
+    navController: NavHostController,
+    mealViewModel: MealViewModel
+) {
   navigation(startDestination = Route.Home, route = Route.Overview) {
     composable(Route.Home) {
       GenericScreen(
           navController = navController,
-          content = { paddingValues -> OverviewScreen(paddingValues) })
+          content = { paddingValues ->
+            OverviewScreen(paddingValues, navController, mealViewModel)
```
sharif-bbd, line 21: instead of passing the navcontroller as argument pass a function of type ()->Unit with navController.navigate(Route.addMeal)

app/src/main/java/com/github/se/polyfit/ui/screen/OverviewScreen.kt
```diff
@@ -198,6 +206,12 @@ fun OverviewScreen(paddingValues: PaddingValues) {
         ->
         val bitmap = result.data?.extras?.get("data") as? Bitmap
         imageBitmap = bitmap
+
+        // observe the live data and log the result on changes
+        SpoonacularApiCaller().getMealsFromImage(imageBitmap!!).observeForever {
+          mealViewModel.setMealData(it)
+          navController.navigate(Route.AddMeal)
```
sharif-bbd, line 213: same here

app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt
RandomUsername1315: The coroutine will shut down because the functions launched within it are guaranteed to end at some point. For example, if the API does not respond within a certain amount of time, it will trigger a timeout, causing every sub-function to return. The issue is that Spoonacular is not run within a single ViewModel. For instance, if it is launched in one ViewModel and that ViewModel is stopped, then its coroutine will also stop. That is why I resorted to a global scope to avoid having any issues if anything happened with the viewmodel.

But creating my own scope could be an idea. I will need to look more into it.

app/src/main/java/com/github/se/polyfit/ui/navigation/NavigationGraph.kt
RandomUsername1315: The overview screen needs to navigate between different screens at the push of a button. Since there are multiple screens you can navigate to, a single function with no arguments cannot perform this task.

PR #112, opened 2024-04-12
Test/end picture

antomax33:
Adding an end to end test. from overview to end


Neb1010 (approved):
LGTM, but maybe for end to end test to have a better "meal" with at least some nutritional informations

PR #117, opened 2024-04-14
FIX: Ensuring no negative nutrients in NutritionalInfo

medwardson:
https://github.com/swent-group10/polyfit/issues/95

Quick PR to prevent Nutritional Information from having negative values. I decided it would be ok to leave the `Nutrient` if it is zero because it is helpful to see that there is no amount of a nutrient visually.

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/model/nutritionalInformation/NutritionalInformation.kt
+++ b/app/src/main/java/com/github/se/polyfit/model/nutritionalInformation/NutritionalInformation.kt
@@ -40,7 +40,8 @@ class NutritionalInformation {
 
   /**
    * Update the nutritional information with new values. If the nutrient type already exists, the
-   * amount is updated. If the nutrient type does not exist, a new nutrient is added.
+   * amount is updated. If the nutrient type does not exist, a new nutrient is added. We make sure
+   * all nutrients have positive amounts.
    *
    * @param newValues the new values to update the nutritional information with
    */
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/polyfit/model/nutritionalInformation/NutritionalInformation.kt
```diff
@@ -53,10 +54,12 @@ class NutritionalInformation {
             val index = nutrients.indexOfFirst { it.nutrientType == newValues.nutrientType }
 
             nutrients[index] = (nutrients[index] + newValues)!!
-          } else {
+          } else if (newValues.amount >= 0.0) {
             nutrients.add(newValues.deepCopy())
           }
         }
+
+    nutrients.removeIf { it.amount < 0.0 }
```
RandomUsername1315, line 62: Do you think it would be a good idea to Log.d the case when this condition occurs? This way, we can track it in case a bug arises elsewhere. Or do you think that it is an necessary ?

RandomUsername1315 (approved):
Nice and short. Awesome !!!

app/src/main/java/com/github/se/polyfit/model/nutritionalInformation/NutritionalInformation.kt
medwardson: I think that's a good idea :-) Will add that in

Issue #116, opened 2024-04-14
Cleanup console warnings


medwardson:
Current warnings in console that should be cleaned:
```
> Task :app:compileDebugAndroidTestKotlin
w: file:///Users/meshaedwardson/uw/classes/4A/CS%20311/polyfit/app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCallerTest.kt:120:12 Condition 'response != null' is always 'true'
w: file:///Users/meshaedwardson/uw/classes/4A/CS%20311/polyfit/app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCallerTest.kt:136:11 Variable 'response' is never used
w: file:///Users/meshaedwardson/uw/classes/4A/CS%20311/polyfit/app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCallerTest.kt:145:12 Condition 'response != null' is always 'true'
w: file:///Users/meshaedwardson/uw/classes/4A/CS%20311/polyfit/app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCallerTest.kt:158:11 Variable 'response' is never used
w: file:///Users/meshaedwardson/uw/classes/4A/CS%20311/polyfit/app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCallerTest.kt:166:12 Condition 'response != null' is always 'true'
w: file:///Users/meshaedwardson/uw/classes/4A/CS%20311/polyfit/app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCallerTest.kt:180:12 Condition 'response != null' is always 'true'
w: file:///Users/meshaedwardson/uw/classes/4A/CS%20311/polyfit/app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/MealFirebaseRepositoryTest.kt:81:9 Variable 'exception' is never used
w: file:///Users/meshaedwardson/uw/classes/4A/CS%20311/polyfit/app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/MealFirebaseRepositoryTest.kt:107:9 Variable 'exception' is never used
w: file:///Users/meshaedwardson/uw/classes/4A/CS%20311/polyfit/app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/MealFirebaseRepositoryTest.kt:136:9 Variable 'result' is never used
[...]

PR #118, opened 2024-04-14
FIX: Console Warnings

medwardson:
https://github.com/swent-group10/polyfit/issues/116

There are a handful of warnings in our console when running tests (see issue above for exact). Cleaning this warnings up now so that our logs are clean and we can easily detect real errors.

### DIFF
--- a/app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/MealFirebaseRepositoryTest.kt
+++ b/app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/MealFirebaseRepositoryTest.kt
@@ -74,12 +74,11 @@ class MealFirebaseRepositoryTest {
   }
 
   @Test
-  fun storeMealFailure() = runBlocking {
+  fun storeMealFailure(): Unit = runBlocking {
     every { mockCollection.document(any()) } returns mockDocumentReference
     every { mockDocumentReference.set(any()) } returns
         Tasks.forException(Exception("Failed to store meal"))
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/MealFirebaseRepositoryTest.kt
```diff
@@ -96,19 +95,17 @@ class MealFirebaseRepositoryTest {
...
         Tasks.forException(Exception("Failed to store meal"))
     every { mockDocumentReference.id } returns "1"
 
-    val exception =
-        assertFailsWith<Exception> {
-          runBlocking { mealFirebaseRepository.storeMeal(mealNoId).await() }
-        }
-
+    assertFailsWith<Exception> {
+      runBlocking { mealFirebaseRepository.storeMeal(mealNoId).await() }
```
RandomUsername1315, line 107: Can't the runBlocking be removed here as it is already in a runBlocking block ?

```suggestion
       mealFirebaseRepository.storeMeal(mealNoId).await() 
```

RandomUsername1315 (approved):
LGTM thank you for fixing the console writing and keeping the PR short :+1:

app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/MealFirebaseRepositoryTest.kt
medwardson: I'm not sure :p I'll check

PR #124, opened 2024-04-16
Creation of local database for meal storage

RandomUsername1315:
# Pull Request Overview

This pull request adds a simple local room database that will be used to store meals offline. The overview model will use the data stored in this database to produce graphs.

## Changes

- **Added `MealDao.kt` class**: This class contains all the potential SQL commands that can be run.
- **Added `MealEntity.kt` Data Classes**: This represents the structure of a meal stored.
- **Added `MealDatabase.kt` Classes**: The actual database where the meal table is stored.
- **Added Tests for `MealDatabase`**.

## Implementation Checklist

- [x] MealDao
- [x] MealEntity
- [x] MealDatabase

## Tests Checklist

- [x] Correct deserialization and conversion from MealEntity to Meal.
- [x] Ensure the commands work correctly and return the correct arguments.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -11,6 +11,8 @@ plugins {
     kotlin("kapt")
     id("com.ncorti.ktfmt.gradle") version "0.16.0"
     id("jacoco")
+    id("kotlin-kapt")
+
 
 }
 
@@ -101,19 +103,15 @@ android {
 
 
     dependencies {
-        implementation("com.squareup.okhttp3:okhttp:4.12.0")
-        androidTestImplementation("com.squareup.okhttp3:okhttp:4.12.0")
-        implementation("com.squareup.okhttp3:okhttp:4.12.0")
-
-        implementation("androidx.core:core-ktx:1.12.0")
+        implementation("androidx.core:core:1.12.0")
         implementation("androidx.lifecycle:lifecycle-runtime-ktx:2.7.0")
         implementation("androidx.activity:activity-compose:1.8.2")
         implementation(platform("androidx.compose:compose-bom:2024.04.00"))
-        implementation("androidx.compose.runtime:runtime-livedata:1.1.0")
+        implementation("androidx.compose.runtime:runtime-livedata:1.6.5")
         implementation("androidx.compose.ui:ui")
         implementation("androidx.compose.ui:ui-graphics")
         implementation("androidx.compose.ui:ui-tooling-preview")
-        implementation("androidx.compose.material3:material3")
+        implementation("androidx.compose.material3:material3:1.2.1")
         implementation("androidx.navigation:navigation-compose:2.7.7")
         implementation("com.google.android.play:core-ktx:1.8.1")
         implementation("org.json:json:20210307")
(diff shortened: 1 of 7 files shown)


medwardson (approved):
It looks good to me charles. 

Two thoughts
1. You may have already planned this and IDK, but maybe we can use this as a way to clean up our MealViewModel? For example, take the picture and then construct and store that meal locally, and then make MealViewModel accept an ID to look up in the localDB? Then we can just redirect to that path and pass a string, should be simple
2. I guess this depends on how we plan on using the local db, but, it might be worth while to create an Ingredient DAO/Database/Entity for faster lookup times there. Specifically if we end up using the ingredients in any of the graphs. I think it would be more efficient that going through the JSON every time.

RandomUsername1315:
@medwardson
Sorry to ping you, I couldn't directly respond to your comment.

1. Yeah, that is a great idea. I was hoping that I could get Hilt to work for an even cleaner implementation. Because the way we do it right now (manually passing the meal viewmodel to the overview and the meal flow) is not really pretty. But if I can't get the hilt testing to work properly, then yeah, we can take this approach.
2. Right now, the issue with the way it is implemented is that if a meal is deleted, then so are the ingredients that it is composed of. I just noticed another issue with this implementation. If we have two meals that contain the same ingredient, the 'get all ingredients' function will return that ingredient twice. What do you mean by the json? Translating from ingredients to a json to be stored in the meal database? Are you suggesting that instead of storing the ingredients in the meal database, we have a dedicated one just for ingredients? (We would need to find a way to link ingredients in their database to their corresponding meal.)

medwardson:
> @medwardson Sorry to ping you, I couldn't directly respond to your comment.
> 
> 1. Yeah, that is a great idea. I was hoping that I could get Hilt to work for an even cleaner implementation. Because the way we do it right now (manually passing the meal viewmodel to the overview and the meal flow) is not really pretty. But if I can't get the hilt testing to work properly, then yeah, we can take this approach.
> 2. Right now, the issue with the way it is implemented is that if a meal is deleted, then so are the ingredients that it is composed of. I just noticed another issue with this implementation. If we have two meals that contain the same ingredient, the 'get all ingredients' function will return that ingredient twice. What do you mean by the json? Translating from ingredients to a json to be stored in the meal database? Are you suggesting that instead of storing the ingredients in the meal database, we have a dedicated one just for ingredients? (We would need to find a way to link ingredients in their database to their corresponding meal.)

1. Sounds good!
2. Yeah, I was suggesting another table for only ingredient, but after reflecting I realize that our ingredients have amount built in, so reusing it will be complex. By JSON i meant: 
```
  fun fromIngredientList(ingredients: List<Ingredient>): String {
    return Gson().toJson(ingredients)
  }

  @TypeConverter
  fun toIngredientList(ingredientsString: String): List<Ingredient> {
    val type = object : TypeToken<List<Ingredient>>() {}.type
    return Gson().fromJson(ingredientsString, type)
  }
```
bc i think this is less efficient than getting from a table, but, i think as mentioned above its not worth our efforts right now so all good :^)

PR #126, opened 2024-04-17
Modify the MealRepo to use the local DB 

RandomUsername1315:
# Pull Request Overview

This is a simple PR that modifies the previous implementation of `MealRepo`. Previously, it relied solely on the Firebase database for storing meals. The new implementation favors the local database and only updates Firebase when possible (i.e., when the device is connected to the internet).

## Changes

- **Modify `MealRepository.kt` Class**: Modified previous implementation to use local DB.
- **Modify `MealDao.kt` Class**: To be able to handle queries needed by `MealRepo`. (Fixed a potential bug that could, in theory, return duplicate ingredients when the `getAllIngredients` function in `MealRepo` was called.)
- **Modify the Tests for `MealDatabase` and `MealRepository`**

## Implementation Checklist

- [x] Modify `MealRepo`
- [x] Fix bug in `MealDao`
- [x] Add new functionality to `MealDao`

## Tests Checklist

- [x] Modify `MealRepository` test to test both online and offline capabilities
- [x] Change the `MealDatabaseTest` to test the new required functionalities

## Questions for PR Review

- Is the way data is only fetched from the cloud when needed okay? Or would it be better to always fetch it from the cloud ?


antomax33 (approved):
LGTM

PR #127, opened 2024-04-17
Add Meal Tags

aprilslis:
We have decided to add tags to each meal creation to put labels like "pet-friendly" or "nut allergy warning" to the meals we create. This PR created a new data class called `MealTag` and the `Meal` data class has a new field that records the tags it has.

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/model/meal/MealTag.kt
+++ b/app/src/main/java/com/github/se/polyfit/model/meal/MealTag.kt
@@ -1,0 +1,62 @@
+package com.github.se.polyfit.model.meal
+
+import android.util.Log
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toArgb
+
+data class MealTag(var tagName: String, var tagColor: Color) {
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (other !is MealTag) return false
+
+    if (tagName != other.tagName) return false
+    if (tagColor != other.tagColor) return false
+
+    return true
+  }
+
+  override fun hashCode(): Int {
+    var result = tagName.hashCode()
+    result = 31 * result + tagColor.hashCode()
+
+    return result
+  }
+
+  fun isComplete(): Boolean {
+    return tagName.isNotEmpty() && tagColor != Color.Unspecified
+  }
+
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/polyfit/model/meal/MealTag.kt
```diff
@@ -0,0 +1,62 @@
+package com.github.se.polyfit.model.meal
+
+import android.util.Log
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toArgb
+
+data class MealTag(var tagName: String, var tagColor: Color) {
```
RandomUsername1315, line 7: It might be better to use an enum instead of a String for the tagName. This would ensure uniformity and potentially prevent errors related to string handling.

RandomUsername1315 (approved):
Really clean implementation! 
I'm not sure if it should be addressed in this PR or a separate one, but given the changes to the meal structure, we should also update the classes that handle data storage on both the Firebase database and the local database.

app/src/main/java/com/github/se/polyfit/model/meal/MealTag.kt
aprilslis: The reason why I used String for the tagName is for allowing user to create their own custom tags, which could be anything. Do you think we could keep string for tagname or just don't allow users to create tags?

aprilslis: I will update the local database but also isnt the firebase database just the serialization and deserialization?

RandomUsername1315: Yes, just both of those functions need to be updated.

aprilslis: i think I already did it for both MealTag and Meal, am I missing anywhere else for the serialization?

RandomUsername1315 (approved):
Everything looks good

app/src/main/java/com/github/se/polyfit/model/meal/MealTag.kt
RandomUsername1315: No that is all. Everything looks good !

PR #129, opened 2024-04-18
Modify mealviewmodel to use mealRepo instead of firebaseMealRepo

RandomUsername1315:
# Pull Request Overview

This simple pull request modifies the `MealViewModel` to use `MealRepository` instead of `FirebaseMealRepository`. This change ensures that data is cached in the local database and fetched from the Firebase database only when necessary.

## Changes

- **Modify `MealViewModel.kt` Class**: Updated the previous implementation to use the repository, which decides whether a request should be made locally or to the Firebase database.
- **Modify `MainActivity.kt` and `AddMealFlow.kt` Classes**: Updated to handle the new implementation of the ViewModel.

## Notes

- The implementation is currently a bit messy, particularly in `MainActivity`. PR  #138 will address this by further integrating Hilt dependency injection, thereby improving the clarity of the code.


app/build.gradle.kts
```diff
@@ -211,3 +211,6 @@ tasks.register("jacocoTestReport", JacocoReport::class) {
         include("outputs/code_coverage/debugAndroidTest/connected/*/coverage.ec")
     })
 }
+dependencies {
+    implementation("androidx.test:core-ktx:1.5.0")
```
antomax33, line 215: Put it with the other implementations

app/src/androidTest/java/com/github/se/polyfit/meal/MealViewModelTest.kt
```diff
@@ -0,0 +1,148 @@
...
+    viewModel = MealViewModel(mealRepo, Meal.default(), "firebaseID")
+  }
+
+  @Test
+  fun setMealName_updates_meal_name_and_completeness() = runBlocking {
+    val mealName = "Test Meal"
+    viewModel.setMealData(Meal.default())
+    viewModel.setMealName(mealName)
+
+    val updatedMeal = viewModel.meal.getOrAwait()
```
antomax33, line 47: Maybe rename getOrAwait to waitData

antomax33 (changes requested):
Just change the place of the implementation in the gradle

app/build.gradle.kts
RandomUsername1315: Thank you for finding this. Sometimes, when you click 'auto import dependency', it for some reason adds it to the bottom of the Gradle file.

app/src/androidTest/java/com/github/se/polyfit/meal/MealViewModelTest.kt
RandomUsername1315: The naming comes from the fact that if the data is already available, there is no need to wait; it directly returns. If the live data has not been updated, then it waits for a certain amount of time (3 seconds in this case). If the data has not changed while waiting, then throw an error.

antomax33 (approved):

PR #130, opened 2024-04-18
Additional Meal Info UI

medwardson:
PR [4/4]

See:
- https://github.com/swent-group10/polyfit/pull/147
- https://github.com/swent-group10/polyfit/pull/148
- https://github.com/swent-group10/polyfit/pull/149

This PR adds a screen for recording additional meal information (date, occasion, and tags). It got added into the AddMealFlow after the ingredients. There were also minor changes to the MealViewModel to support these changes as well as to the Meal data class.

<hr>

_**NOTE**_

JaCoCo was refusing to upload coverage for one of the tests, causing the SonarCloud to say that the file has 0% line coverage. @aprilslis discovered that the issue comes from an error says `java.lang.IllegalStateException: Line 36 is out of range in the file`, on a file that is only 34 lines long. After researching, it was discovered that this is a known issue caused by Jacoco itself. The error occurs because Jacoco generates code for each file to help test the coverages, and then the XML coverage report mistakenly says that the file has more lines than it does, which later causes SonarCloud to be confused and recognize it as an error. Newer versions of Jacoco mostly fix this issue, but in our case, it did not. Thus as a workaround, we are adding comments to make line 36 exist, it then properly updates coverage. Currently, no other fixes worked so that's what we'll have to use for now.


app/src/main/java/com/github/se/polyfit/ui/screen/AdditionalMealInfoScreen.kt
```diff
@@ -0,0 +1,79 @@
...
+  val isComplete = occasion != MealOccasion.OTHER
+
+  Scaffold(
+      topBar = { SimpleTopBar("Additional Meal Info", navigateBack) },
+      bottomBar = { BottomBar(isComplete, navigateForward) }) {
+        Column(
+            modifier = Modifier.padding(it).fillMaxWidth().testTag("AdditionalMealInfoContainer")) {
+              DateSelector(onConfirm = mealViewModel::setMealCreatedAt)
+              MealOccasionSelector(occasion, mealViewModel::setMealOccasion)
+              MealTagSelector(
```
RandomUsername1315, line 51: ## Issue: UI Scaling with Multiple Tags

**Problem Description:**
There is an  issue with the UI for the `Special Tags` when there is a large number of tags. Currently, adding many tags causes the bottom tabs to become inaccessible, this could be due an overflow issue with the tag container.

**Screenshot:**
Screenshot of UI Issue

**Proposed Solutions:**
**Scrollable Tag Container:** Implement scrolling for the tags area to keep the bottom tabs accessible, no matter how many tags are displayed/added.

RandomUsername1315 (approved):
Amazing PR ! Small issue with the UI for the tags. (Could be fixed potentially later)

app/src/main/java/com/github/se/polyfit/ui/screen/AdditionalMealInfoScreen.kt
medwardson: I'm wondering if we should just limit the amount of tags: i.e. 5, 10. At a certain point its a bit redundant and someone could fill up our DB with a bunch of trash

medwardson: If that works with you, I'll leave it as a ticket in the back log for the future. Gonna try to shift focus towards the day summary UI

medwardson: actually ill just add the limit in now

PR #133, opened 2024-04-18
Atttemps to get a cleaner implementation by using Hilt DI

RandomUsername1315:


RandomUsername1315:
Issue with branch, will create a new branch and PR

PR #139, opened 2024-04-19
Fix/register

antomax33:


app/src/androidTest/java/com/github/se/polyfit/ui/screen/LoginScreen.kt
```diff
@@ -9,7 +9,7 @@ class LoginScreen(semanticsProvider: SemanticsNodeInteractionsProvider) :
         semanticsProvider = semanticsProvider, viewBuilderAction = { hasTestTag("LoginScreen") }) {
 
   // Structural elements of the UI
-  val loginTitle: KNode = child { hasTestTag("LoginTitle") }
+  val loginTitle: KNode = child { hasTestTag("TopBarTitle") }
```
RandomUsername1315, line 12: Good find. I think this bug occurred because at some point we had both a login and a sign-in screen, which led to confusion.

RandomUsername1315 (approved):
Thank you for finding and fixing this very annoying bug !

PR #140, opened 2024-04-19
Representing and Storing a post locally and on firebase database

RandomUsername1315:
Sets up the basic structure of what data a post should contain.

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/model/post/Post.kt
+++ b/app/src/main/java/com/github/se/polyfit/model/post/Post.kt
@@ -1,0 +1,53 @@
+package com.github.se.polyfit.model.post
+
+import android.graphics.Bitmap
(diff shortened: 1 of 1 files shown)


RandomUsername1315:
@aprilslis, looking at the Figma, this is what I came up with for what is needed for the 'Create a Post' screen. What do you think? I am not so sure about storing photos. Was it a placeholder or is it something that you really want there to be?

aprilslis:
> @aprilslis, looking at the Figma, this is what I came up with for what is needed for the 'Create a Post' screen. What do you think? I am not so sure about storing photos. Was it a placeholder or is it something that you really want there to be?

Conversation discussed offline.

aprilslis (approved):
Looks great overall!

PR #141, opened 2024-04-20
Fix/modify mealrepo to use locald db

RandomUsername1315:
# Pull Request Overview

This simple pull request modifies the `MealViewModel` to use `MealRepository` instead of `FirebaseMealRepository`. This change ensures that data is cached in the local database and fetched from the Firebase database only when necessary.

## Changes

- **Modify `MealViewModel.kt` Class**: Updated the previous implementation to use the repository, which decides whether a request should be made locally or to the Firebase database.
- **Modify `MainActivity.kt` and `AddMealFlow.kt` Classes**: Updated to handle the new implementation of the ViewModel.

## Notes

- #126 was approved but did not merge correctly with the main branch. When #114 was approved for merging, it did not directly target the main branch; instead, it targeted the branch that created the local database for meals (which had been merged first). When #126 was approved and merged, the other branch was reinstated. Causing the need for this PR, even if all changes were already approved. 
- This change also makes use of Hilt Di for a cleaner implementation


app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
```diff
@@ -30,30 +32,70 @@ constructor(private val user: User, private val mealRepo: MealFirebaseRepository
...
+            mealTemp,
+            nutritionalInformation,
+            ingredients,
+            firebaseID,
+            createdAt)
+
+    _isComplete.value = _meal.value?.isComplete() ?: false
   }
 
   fun setMealName(name: String) {
```
aprilslis, line 93: Now that we have `setMealData`, `setMealName` is not really needed, we should maybe remove it and just use `setMealData` for all name-changing place

app/src/androidTest/java/com/github/se/polyfit/viewmodel/MealViewModelTest.kt
```diff
@@ -0,0 +1,224 @@
...
+    io.mockk.clearMocks(mealRepo)
+  }
+
+  @Test
+  fun testInitialization_withFirebaseID_loadsData() {
+    assert(viewModel.meal.value?.name == "Test Meal")
+  }
+
+  @Test
+  fun testSetMealName_updatesMealName() {
```
aprilslis, line 68: this test will no longer be needed if you remove `setMealName` and use `setMealData` only

app/src/androidTest/java/com/github/se/polyfit/MainActivityTest.kt
```diff
@@ -0,0 +1,19 @@
...
+import androidx.test.filters.LargeTest
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+@LargeTest
+class MainActivityTest {
+
+  @Test
+  fun testActivityShouldBeLaunched() {
```
aprilslis, line 14: Why is this empty test here? MainActivity is not changed, nor the test is testing for anything?

aprilslis (approved):
Looks overall! after you resolve conflicts and comment it would be good!

app/src/androidTest/java/com/github/se/polyfit/MainActivityTest.kt
RandomUsername1315: Haha, I don't know how that got there. It's definitely a mistake. Thanks for catching it.

PR #143, opened 2024-04-20
Create Meal without a Photo

medwardson:
Allow users to create a meal without taking a picture. 

I took the MealViewModel out of the main activity at least and brought it to our main nav graph as it didn't need to be such a high level. Ideally when we can pass around ids rather than mealViewModel, we can get rid of the need to reset it.

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
+++ b/app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
@@ -96,4 +96,10 @@ class MealViewModel @Inject constructor(private val mealRepo: MealRepository) : 
   fun removeTag(tag: MealTag) {
     _meal.value!!.tags.remove(tag)
   }
+
+  // TODO: This can be removed once we are properly using a new ViewModel for each Meal
+  fun reset() {
+    _meal.value = Meal.default()
+    _isComplete.value = false
+  }
 }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/se/polyfit/MainActivity.kt
```diff
@@ -35,23 +33,13 @@ class MainActivity : ComponentActivity() {
...
         NavHost(navController = navController, startDestination = Route.Register) {
-          globalNavigation(navController, mealViewModel)
+          globalNavigation(navController)
           composable(Route.Register) { LoginScreen(navigation::navigateToHome) }
-
-          composable(Route.AddMeal) {
-            // make sure the create is clear
-
-            // check reall created
-            AddMealFlow(navigation::goBack, navigation::navigateToHome, "testUserID", mealViewModel)
```
antomax33, line 53: :+1:

antomax33 (approved):
LGTM

app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
```diff
@@ -84,4 +84,10 @@ class MealViewModel(
     }
     _isComplete.value = _meal.value?.isComplete() ?: false
   }
+
+  // TODO: This can be removed once we are properly using a new ViewModel for each Meal
```
RandomUsername1315, line 88: I think that we have completely different views on what the lifespan of a `MealViewModel` should be. The way I see it, we only have one, but it seems you view this as creating a new one for each meal creation. Let's talk tomorrow after the stand-up meeting to make sure we are on the same page about this.

aprilslis (approved):
LGTM!

PR #144, opened 2024-04-20
Add folder structure for UI components

medwardson:
Our components are getting to be a long list, adding some more folders to segment and make it easier to find files.


RandomUsername1315 (approved):
So many important changes, haha. Thank you for cleaning up the project a bit.

PR #145, opened 2024-04-20
Feature/UI/map

antomax33:
## Map
We can display posts on the map from where they have been sent, within a certain radius ( 100m-2km)
image

### Persons that worked on this PR
- Benjamin : change on `MainActivity` , `BottomNavigation` , `GenericScreen` , `LauchedEffect` in the `MapScreen` and `LaucherForActivityResult`  
- Antoine : Worked on the UI elements (ex: the map, the sliders and the buttons)
- Charles :  The backend storing posts and the displaying the markers at the correct location.


app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
```diff
@@ -89,6 +90,8 @@ class PostFirebaseRepository(
       completion: (List<Post>) -> Unit,
       geoFire: GeoFire = GeoFire(geoFireRef)
   ) {
+    Log.d("MapRadius", "Radius: $radiusInKm")
+    Log.d("MapCenter", "Center: $centerLatitude, $centerLongitude")
```
RandomUsername1315, line 94: To avoid congesting the logs maybe a good idea to remove the debug: 
```suggestion

```

Same thing same for the other `log` in this file

RandomUsername1315 (approved):
Looks awesome !

Neb1010:
Map User story done 👍

PR #147, opened 2024-04-20
Meal Occasion Selector

medwardson:
PR [1/n]

Selector to choose a meal occasion and update the value using a callback. First of a few PRs for the additional info UI to keep things small.

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/ui/components/selector/MealOccasionSelector.kt
+++ b/app/src/main/java/com/github/se/polyfit/ui/components/selector/MealOccasionSelector.kt
@@ -1,0 +1,96 @@
+package com.github.se.polyfit.ui.components.selector
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.RadioButton
+import androidx.compose.material3.RadioButtonColors
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.github.se.polyfit.model.meal.MealOccasion
+import com.github.se.polyfit.ui.theme.PrimaryPurple
+import com.github.se.polyfit.ui.theme.PurpleGrey40
+import com.github.se.polyfit.ui.utils.titleCase
+
+@Composable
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/se/polyfit/ui/components/selector/MealOccasionSelector.kt
```diff
@@ -0,0 +1,96 @@
...
+import androidx.compose.ui.unit.dp
+import com.github.se.polyfit.model.meal.MealOccasion
+import com.github.se.polyfit.ui.theme.PrimaryPurple
+import com.github.se.polyfit.ui.theme.PurpleGrey40
+import com.github.se.polyfit.ui.utils.titleCase
+
+@Composable
+fun MealOccasionSelector(
+    startOccasion: MealOccasion,
+    onConfirm: (MealOccasion) -> Unit,
```
RandomUsername1315, line 30: Code looks amazing ! If I  understood correctly the `onConfirm` function, will update the meal in the `mealViewModel` every time the user clicks on a new occasion ?

app/src/main/java/com/github/se/polyfit/ui/components/selector/MealOccasionSelector.kt
```diff
@@ -0,0 +1,96 @@
...
+  Column(modifier = modifier.fillMaxWidth().testTag("MealOccasionSelector")) {
+    Text(
+        text = "Meal Occasion",
+        style = MaterialTheme.typography.titleLarge,
+        color = PurpleGrey40,
+        modifier = Modifier.padding(16.dp, 0.dp).testTag("Title"))
+    Row(
+        modifier = Modifier.fillMaxWidth().testTag("ButtonRow"),
+        horizontalArrangement = Arrangement.SpaceEvenly) {
+          Column(modifier = Modifier.testTag("LeftColumn")) {
```
RandomUsername1315, line 49: It might be a good idea to iterate over the elements of the enum instead of using hardcoded values. This could improve modularity if we decide to add or remove a MealOccasion in the future.

RandomUsername1315 (approved):
Thank you amazing short pr !

app/src/main/java/com/github/se/polyfit/ui/components/selector/MealOccasionSelector.kt
medwardson: That's what it'll be used for yeah :D But to make it flexible, instead of having to pass a mealviewmodel you can give it any callback to update whatever you want

app/src/main/java/com/github/se/polyfit/ui/components/selector/MealOccasionSelector.kt
medwardson: It's a good point. I think as a counter argument, two things:
1. I think its highly unlikely for the enum to change. At worst the name would change, and the code wont compile and will need a quick change. I think there will only ever really be 4 options for this.
2. We have an "Other" enum, but that's basically just used to create a new meal and have it as an invalid state until a proper one is selected. If we iterate over the enums, we would have to hardcode to exclude it and this would become a burden in a different way if we change the order/add to the enums.

RandomUsername1315: I missed the fact that you did not include `MealOccasion.OTHER`; yes, your implementation is better suited to what I was thinking.

PR #148, opened 2024-04-20
Date Selector

medwardson:
PR [2/n]

Selector to choose a date and update the value using a callback. Second of a few PRs for the additional info UI to keep things small.

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
+++ b/app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
@@ -1,0 +1,111 @@
+package com.github.se.polyfit.ui.components.selector
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.DatePicker
+import androidx.compose.material3.DatePickerDialog
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.material3.rememberDatePickerState
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.style.TextDecoration
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.github.se.polyfit.ui.theme.PrimaryPink
+import com.github.se.polyfit.ui.theme.PurpleGrey40
+import java.time.LocalDate
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DateSelector(onConfirm: (LocalDate) -> Unit, modifier: Modifier = Modifier) {
+  var showDatePicker by remember { mutableStateOf(false) }
+
+  // I'm *pretty sure* this should select the right time and zone, but if test flakes this is why
+  val initialMillis = LocalDate.now(ZoneId.systemDefault()).toEpochDay() * 1000 * 60 * 60 * 24
+  val datePickerState = rememberDatePickerState(initialSelectedDateMillis = initialMillis)
+
+  val selectedDate =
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
```diff
@@ -0,0 +1,110 @@
...
+                modifier = Modifier.testTag("DatePickerDialog")) {
+                  DatePicker(state = datePickerState, modifier = Modifier.testTag("DatePicker"))
+                }
+          }
+        }
+  }
+}
+
+@Composable
+private fun FormatDate(dateString: String) {
```
RandomUsername1315, line 81: ## Bug Report: App Crash on Date Input Manipulation

**Issue Summary:**  
The application crashes when a user attempts to delete part of the year component (or any parts of the date). This was observed when modifying the date "12/03/2024" to remove the last digit of the year.

**Steps to Reproduce:**
1. Open the date entry field.
2. Click on the pencil icon to input the data with the keyboard instead of the gui.
3. Delete the digit '4' from the year.

**Expected Result:**  
Disable the done button until the user entered a correctly formatted date.

**Actual Result:**  
The preview crashes.

**Video Proof:**  
[Screen_recording_20240421_111431.webm](https://github.com/swent-group10/polyfit/assets/100274930/0ce2f805-ad30-4adb-b441-f3d4506f69d0)


---

app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
```diff
@@ -0,0 +1,110 @@
...
+  val day = dateString.substring(8, 10)
+  val month = dateString.substring(5, 7)
+  val year = dateString.substring(0, 4)
```
RandomUsername1315, line 84: The `.substring` is culprit for the preview crashes on wrongly formatted user date input.

app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
```diff
@@ -0,0 +1,110 @@
...
+          TextButton(
+              onClick = { showDatePicker = true },
+              modifier = Modifier.testTag("ChangeDateButton")) {
+                FormatDate(dateString)
+              }
+
+          if (showDatePicker) {
+            DatePickerDialog(
+                onDismissRequest = { showDatePicker = false },
+                confirmButton = { TextButton(onClick = { onConfirm() }) { Text(text = "Done") } },
```
RandomUsername1315, line 68: It would be a good idea to disable the button until a validly formatted date is entered.

app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
```diff
@@ -0,0 +1,110 @@
...
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DateSelector(onConfirm: (LocalDate) -> Unit, modifier: Modifier = Modifier) {
+  var showDatePicker by remember { mutableStateOf(false) }
+
+  // I'm *pretty sure* this should select the right time and zone, but if test flakes this is why
+  val initialMillis = LocalDate.now().atStartOfDay().toInstant(ZoneOffset.UTC).toEpochMilli()
```
RandomUsername1315, line 37: For example, if the user is in a time zone that is a day behind UTC, this would cause the "initial start date" (`initialMillis`) to be incorrect. However, this issue is relatively minor.

RandomUsername1315 (changes requested):
Looks good, just a small bug that causes the preview to crash.

app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
medwardson: i think im just gonna remove writing it by keyboard, i didn't intend that to be a feature and it causes more issues than benefit ;P

app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
medwardson: No longer an issue with removing typing, will always be a valid date. I'll change it to use date formatting instead of string splitting

app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
medwardson: Changed to use device time, which, for some reason is giving me the wrong date, but i think it may be due to my computer having north America setting

app/src/main/java/com/github/se/polyfit/ui/components/selector/DateSelector.kt
medwardson: Good find 🚀 Now users have to select a date from the dialog and can't input any invalid date

RandomUsername1315 (approved):
Everything looks good !

PR #155, opened 2024-04-23
Create post viewmodel 

RandomUsername1315:
# Pull Request Overview

- This PR builds on the previous one (#136) by creating a ViewModel that will be used in the UI interface for creating a meal.

## Changes

- **Create `CreatePostViewModel.kt` Class**: Handles processing the data needed for the UI used to create a new post.

## Implementation Checklist

- [x] Create ViewModel class.

## Tests Checklist

- [x] ViewModel testing.


RandomUsername1315:
@aprilslis would love to get your feedback on this. Is this ok, or do you need any other data for the create a post UI ?

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
```diff
@@ -0,0 +1,71 @@
...
+constructor(
+    private val mealRepository: MealRepository,
+    private val postFirebaseRepository: PostFirebaseRepository,
+) : ViewModel() {
+  private val _post = Post.default()
+
+  val post: UnmodifiablePost
+    get() = _post
+
+  suspend fun getRecentMeals() = mealRepository.getAllMeals().sortedBy { it.createdAt }
```
aprilslis, line 27: nit: Getting all the meals from a user could potentially be expensive and only getting meals within a certain amount of time could be better. That being said, for the scope of our project, it's probably fine.

app/src/main/java/com/github/se/polyfit/di/UserModule.kt
```diff
@@ -63,6 +64,12 @@ object UserModule {
     return MealRepository(context, mealFirebaseRepository, mealDao)
   }
 
+  @Provides
+  @Singleton
+  fun providePostFirebaseRepository(): PostFirebaseRepository {
```
aprilslis, line 69: It seems like this function is not used anywhere right now, is it for future uses?

aprilslis (approved):
Looks good overall, thanks for putting it up!

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
RandomUsername1315: Discussed offline: to create a new Post, only the last five recent meals will be shown as potential options, instead of all of them.

PR #158, opened 2024-04-24
UI/make a post

aprilslis:
UI, UI test for CreatePostScreen. It is fully integrated with backend. 

On this screen, user should be able to draft a post. They can put in optional text descriptions for their post, and are required to select a meal created by themselves to share a post. Before the post is submitted, users are asked for permission to use their location, as the post is visible to those around them.

(Note: see #187 for details on individual components)

image

image

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/data/repository/MealRepository.kt
+++ b/app/src/main/java/com/github/se/polyfit/data/repository/MealRepository.kt
@@ -64,7 +64,7 @@ class MealRepository(
   }
 
   suspend fun getAllMeals(): List<Meal> {
-    return mealDao.getAllMeals()
+    return withContext(this.dispatcher) { mealDao.getAllMeals() }
   }
 
   suspend fun getMealsOnDate(date: LocalDate): List<Meal> {
--- a/app/src/main/java/com/github/se/polyfit/ui/screen/CreatePostScreen.kt
+++ b/app/src/main/java/com/github/se/polyfit/ui/screen/CreatePostScreen.kt
@@ -1,0 +1,161 @@
+package com.github.se.polyfit.ui.screen
+
+import android.util.Log
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.heightIn
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Edit
+import androidx.compose.material3.HorizontalDivider
+import androidx.compose.material3.Icon
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Scaffold
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextField
+import androidx.compose.material3.TextFieldDefaults
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.derivedStateOf
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import androidx.hilt.navigation.compose.hiltViewModel
+import com.github.se.polyfit.R
+import com.github.se.polyfit.model.meal.Meal
+import com.github.se.polyfit.model.post.Location
+import com.github.se.polyfit.ui.components.button.PrimaryButton
+import com.github.se.polyfit.ui.components.dialog.LocationPermissionDialog
+import com.github.se.polyfit.ui.components.scaffold.CenteredTopBar
+import com.github.se.polyfit.ui.components.selector.MealSelector
+import com.github.se.polyfit.ui.components.selector.PictureSelector
+import com.github.se.polyfit.ui.theme.PrimaryPink
+import com.github.se.polyfit.ui.theme.PrimaryPurple
+import com.github.se.polyfit.ui.theme.SecondaryGrey
+import com.github.se.polyfit.viewmodel.post.CreatePostViewModel
+
+@Composable
+fun CreatePostScreen(
+    navigateBack: () -> Unit = {},
+    navigateForward: () -> Unit = {},
+    postViewModel: CreatePostViewModel = hiltViewModel()
+) {
+  val context = LocalContext.current
+  val meals by postViewModel.meals.collectAsState()
+  var isPermissionDialogDisplay by remember { mutableStateOf(false) }
+  var selectedMeal by remember { mutableStateOf(Meal.default()) }
+  val postComplete by remember(selectedMeal) { derivedStateOf { selectedMeal.isComplete() } }
+
+  fun setPostMeal(meal: Meal) {
+    postViewModel.setPostData(meal = meal)
+  }
+
+  fun onApprove() {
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/github/se/polyfit/ui/components/selector/MealSelector.kt
sharif-bbd, line 1: Overall this file contains a lot Logic. A suggestion would be to move as much as you can to a Viewmodel to leave the UI files solely about rendering the display but this isn't really urgent. 
Also make sure to put your hardcoded strings in the String resource file and make sure to use the MaterialTheme whenever possible

settings.gradle.kts
```diff
@@ -13,6 +13,6 @@ dependencyResolutionManagement {
     }
 }
 
-rootProject.name = "polyfit"
+rootProject.name = "Polyfit"
```
sharif-bbd, line 16: This used to cause errors, at least still does on my end

sharif-bbd (changes requested):
Overall the PR is good and well structured, a few changes are to be made but nothing major. Now you also need to add the integration tests. Good PR

app/src/main/java/com/github/se/polyfit/ui/components/selector/MealSelector.kt
RandomUsername1315: @sharif-bbd, great suggestion! I am the one who made the view model for creating a post. What functionality do you think should be moved to the `CreatePostViewModel`?

sharif-bbd: I think the search bar logic although not filtering options at the moment could be moved to a viewmodel and also the completed mealList although temporary could be moved out of the files. And in general fashion the mutablestate variables could be declared in the ViewModel and collected in the UI so instead of setting the value manually in the UI you would call a set function defined in the ViewModel.

settings.gradle.kts
aprilslis: Yes, the PR is still a draft so I haven't prepared it for review

app/src/main/java/com/github/se/polyfit/data/repository/MealRepository.kt
```diff
@@ -64,7 +64,7 @@ class MealRepository(
   }
 
   suspend fun getAllMeals(): List<Meal> {
-    return mealDao.getAllMeals()
+    return withContext(this.dispatcher) { mealDao.getAllMeals() }
```
RandomUsername1315, line 67: Was this causing you an error of the type that running this on the main thread might cause  weird behavior ? If so, in the next iteration of the mealRepo, I will update that for all the functions that need to fetch data from the database.

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
```diff
@@ -60,12 +76,14 @@ constructor(
...
   }
 
-  suspend fun setPost() {
-    try {
-      postFirebaseRepository.storePost(_post)
-    } catch (e: Exception) {
-      Log.e("CreatePostViewModel", "Failed to store post in the database : ${e.message}", e)
-      throw Exception("Failed to store post in the database : ${e.message}")
+  fun setPost() {
+    runBlocking {
```
RandomUsername1315, line 80: OK for now, but to avoid blocking the whole app in the future, we could launch a global coroutine to make the experience more fluid. Or we could just call that technical dept of the v1 that could be fixed in the v2.

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
```diff
@@ -26,6 +32,16 @@ constructor(
 
   suspend fun getRecentMeals() = mealRepository.getAllMeals().sortedBy { it.createdAt }.take(5)
 
+  private val _meals = MutableStateFlow<List<Meal>>(emptyList())
+  val meals = _meals
+
+  init {
+    viewModelScope.launch {
```
RandomUsername1315, line 39: Just need to be careful with using the `viewModelScope`. It caused a previous issue with the `CreateMealViewModel` not being synchronized correctly. But it should be okay as a new `ViewModel` is created each time by Hilt.

RandomUsername1315 (approved):
Just a few remarks. How do we navigate to the create a post screen? Is that set up by another PR?

app/src/main/java/com/github/se/polyfit/ui/screen/CreatePostScreen.kt
```diff
@@ -0,0 +1,161 @@
...
+              onClick = {
+                onButtonPressed()
+                Log.v("Finished", "Clicked")
+              },
+              text = context.getString(R.string.postButton),
+              fontSize = 24,
+              modifier = Modifier.width(200.dp).testTag("PostButton"),
+              color = PrimaryPink,
+              isEnabled = postComplete,
+              buttonShape = RoundedCornerShape(12.dp))
```
sharif-bbd, line 152: Make sure to use the MaterialTheme colors and typography for consistency

app/src/main/java/com/github/se/polyfit/ui/screen/CreatePostScreen.kt
```diff
@@ -0,0 +1,161 @@
...
+            title = context.getString(R.string.newPostTitle), navigateBack = navigateBack)
+      },
+      bottomBar = {
+        BottomBar(
+            postComplete = postComplete, onButtonPressed = { isPermissionDialogDisplay = true })
+      }) {
+        LazyColumn(modifier = Modifier.padding(it).testTag("CreatePostScreen")) {
+          item { PictureSelector(modifier = Modifier.padding(top = 10.dp)) }
+          item { PostDescription(postViewModel::setPostDescription) }
+          item { HorizontalDivider(thickness = 1.dp, color = Color.LightGray) }
```
sharif-bbd, line 86: prioritize the use of MaterialTheme for the colors

sharif-bbd (approved):
Excellent Job ! Very good implementation.

app/src/main/java/com/github/se/polyfit/data/repository/MealRepository.kt
aprilslis: It was yeah, thanks!

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
aprilslis: Yeah, that makes sense. It was to solve issue with threads too

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
aprilslis: Will do!

app/src/main/java/com/github/se/polyfit/ui/screen/CreatePostScreen.kt
aprilslis: As mentioned before, it's best to keep using the variables we have been around the app until we do a full migration.

app/src/main/java/com/github/se/polyfit/ui/screen/CreatePostScreen.kt
```diff
@@ -0,0 +1,161 @@
...
+  fun setPostMeal(meal: Meal) {
+    postViewModel.setPostData(meal = meal)
+  }
+
+  fun onApprove() {
+    isPermissionDialogDisplay = false
+    postViewModel.setPostLocation(Location.default()) // TODO: Include location when ready
+    postViewModel.setPost()
+    navigateForward()
+  }
```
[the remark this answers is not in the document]
aprilslis: This is because the implementation of location is not ready. Once it is, we would only do setPost and navigateForward here, so since its temporary it is not worth making into the view model

app/src/main/java/com/github/se/polyfit/ui/screen/CreatePostScreen.kt
aprilslis: Same as abvove

aprilslis:
@RandomUsername1315 navigation has not been added yet, designs are not yet made but i guess from the home screen

RandomUsername1315:
@aprilslis are we going to add this feature to milestones 2 ?

aprilslis:
@RandomUsername1315 I do want to merge this PR in for milestone 2, but, whether the screen is available for the demo depends on the time we have and if the rest of post feature is ready

RandomUsername1315:
@aprilslis I mean everything is ready, the only thing missing is pictures.

PR #162, opened 2024-04-25
Data Processing

RandomUsername1315:
# Pull Request Overview

This PR introduces a new class dedicated to handling data processing tasks from the local database. By centralizing data operations, this class aims to reduce complexity in the UI and view models. Initially, it will be utilized for generating data for  graphs on the main/overview page and compiling daily calorie summaries.

## Changes Made
- **(Details will be added upon completion)**

## Implementation Checklist

- [x] Update `MealDao` to retrieve all meals for a given day.
- [x] Implement retrieval of average daily calorie data based on date.
- [x] Calculate average calorie intake for the past week and month.

## Testing Checklist
- [x] Verify that `MealDao` operates as expected.
- [x] Ensure that `LocalDataProcessor.kt` returns correct values.

## Questions for Reviewers

- How do you feel about the return type for the `getCaloriesPerMealOccasionToday` function? Do you have suggestions for improvement?
- I'm considering renaming `LocalDataProcessor.kt` as I'm not fully satisfied with the current name. Do you think it adequately reflects the responsibilities of the class or do you have other suggestions?

## Future Improvements

- Extend the functionality to track weight and correlate daily caloric intake with weight changes.


app/src/main/java/com/github/se/polyfit/data/local/dao/MealDao.kt
```diff
@@ -37,6 +38,9 @@ interface MealDao {
     return meal?.toMeal()
   }
 
+  @Query("SELECT * FROM MEALTABLE WHERE createdAt >= :date ")
```
medwardson, line 41: I think we will also need a method to get meals created on a specific date for the daily recap screens, as users will be able to select yesterday for example and see what they ate only that day

app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.github.se.polyfit.data.local.dao.MealDao
+import com.github.se.polyfit.model.meal.MealOccasion
+import java.time.LocalDate
+import javax.inject.Inject
+
+data class DailyCalorieSummary(val date: LocalDate, val totalCalories: Double)
+
+class LocalDataProcessor @Inject constructor(private val mealDao: MealDao) {
+
+  private fun calculateCalories(sinceDate: LocalDate): List<DailyCalorieSummary> {
```
medwardson, line 12: To accomodate the other change suggested above, perhaps this can have an optional parameter to calculate on the date given only or since the date given

medwardson (commented):
It looks good, but I think it needs some adaptation to support single day searching.

> How do you feel about the return type for the getCaloriesPerMealOccasionToday function? Do you have suggestions for improvement?

I think you can return the calories without the date. I think this because if the user is requested the calories since date, they already have that data and won't need it again.

> I'm considering renaming LocalDataProcessor.kt as I'm not fully satisfied with the current name. Do you think it adequately reflects the responsibilities of the class or do you have other suggestions?
I don't see any issues with it personally

app/src/main/java/com/github/se/polyfit/data/local/dao/MealDao.kt
```diff
@@ -37,6 +38,12 @@ interface MealDao {
...
+  fun getMealsCreatedOnOrAfterDate(date: LocalDate): List<Meal>
+
+  @Query("SELECT * FROM MEALTABLE WHERE createdAt <= :date ")
```
medwardson, line 44: i think this should be `createdAt == :date`?

app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
```diff
@@ -0,0 +1,40 @@
...
+        .map { (date, meals) ->
+          DailyCalorieSummary(date, meals.sumOf { it.calculateTotalCalories() })
+        }
+  }
+
+  fun getCaloriesPerMealOccasionToday(): Map<MealOccasion, Double> {
+    return getCaloriesPerMealOnSpecificDay(LocalDate.now())
+  }
+
+  fun getCaloriesPerMealOnSpecificDay(day: LocalDate): Map<MealOccasion, Double> {
```
medwardson, line 24: ty :D

app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import com.github.se.polyfit.data.local.dao.MealDao
+import com.github.se.polyfit.model.meal.MealOccasion
+import java.time.LocalDate
+import javax.inject.Inject
+
+data class DailyCalorieSummary(val date: LocalDate, val totalCalories: Double)
+
+class LocalDataProcessor @Inject constructor(private val mealDao: MealDao) {
+  private fun calculateCalories(sinceDate: LocalDate): List<DailyCalorieSummary> {
```
medwardson, line 11:
```suggestion
  private fun calculateCaloriesSince(date: LocalDate): List<DailyCalorieSummary> {
```
perhaps more clear name

app/src/androidTest/java/com/github/se/polyfit/data/local/database/MealDatabaseTest.kt
```diff
@@ -128,6 +130,51 @@ class MealDatabaseTest {
...
+    assert(newMeals.contains(meal3))
+  }
+
+  private fun createMeal(name: String, firebaseId: String, createdAt: LocalDate): Meal {
+    return Meal(
+        name = name,
+        occasion = MealOccasion.BREAKFAST,
+        mealID = 1,
+        mealTemp = 20.0,
+        nutritionalInformation =
```
medwardson, line 163: i don't think it affects this test, but in general after that bugfix I've learned it is better to create nutrition through your ingredients rather than directly in the meal. perhaps we should even make it not possible to do so

medwardson (approved):
Nicely done

app/src/main/java/com/github/se/polyfit/data/local/dao/MealDao.kt
RandomUsername1315: Thank you for catching that. It is a typo!

app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
RandomUsername1315: That is a much better name. Thank you for the suggestion!

app/src/androidTest/java/com/github/se/polyfit/data/local/database/MealDatabaseTest.kt
RandomUsername1315: I see what you mean. The auto-update feature of meals could cause the recalculation of `nutritionalInformation` to be off from what is expected. We could fix this in a later PR by setting the `nutritionalInformation` to empty by default, to be sure that the correct values are automatically calculated.

PR #165, opened 2024-04-25
Feature/graph screen

sharif-bbd:
### Full Screen Presenting Graph and Data History

This screen displays a graph on the upper half of the screen with weight in ordinates and dates in abscissa. On the lower have you get a list of the data displayed on the graph and the screen allows for searching to filter out the data as well as sorting through KCAL or WEIGHT in Ascending order.

Capture d’écran 2024-04-26 à 01 01 33

Capture d’écran 2024-04-26 à 01 05 57


app/src/main/java/com/github/se/polyfit/ui/viewModel/GraphViewModel.kt
```diff
@@ -0,0 +1,128 @@
...
+              }
+          }
+      }
+      .stateIn(
+          viewModelScope,
+          SharingStarted.WhileSubscribed(5000),
+          _graphData.value
+      )
+
+   */
```
RandomUsername1315, line 81:
```suggestion
  
```

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
```diff
@@ -0,0 +1,99 @@
...
+import com.github.se.polyfit.ui.viewModel.SortPoints
+
+@Composable
+fun FullGraphScreen() {
+  val isTestEnvironment = System.getProperty("isTestEnvironment") == "true"
+  val isExpanded = remember { mutableStateOf(false) }
+
+  val sortedBy = remember { mutableStateOf("KCAL") }
+
+  val viewModel = viewModel<GraphViewModel>()
```
RandomUsername1315, line 42: To keep things consistent across the app, it would be great if you used Hilt DI to inject the ViewModel.

app/src/main/java/com/github/se/polyfit/ui/viewModel/GraphViewModel.kt
```diff
@@ -0,0 +1,128 @@
...
+  private val _sortDirection = MutableStateFlow(SortDirection.ASCENDING)
+
+  private val _searchText = MutableStateFlow("")
+  val searchText = _searchText.asStateFlow()
+
+  private val _isSearching = MutableStateFlow(false)
+  val isSearching = _isSearching.asStateFlow()
+
+  private val _graphData = MutableStateFlow(mockData)
+  val graphData =
```
RandomUsername1315, line 44: Small suggestion to make understanding what this variable mean could be to refactor the sorting into its own function.

RandomUsername1315 (changes requested):
Small changes are needed to improve code consistency and one import is missing that is necessary for the code to compile.

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
```diff
@@ -0,0 +1,99 @@
...
+import androidx.lifecycle.viewmodel.compose.viewModel
+import co.yml.charts.ui.linechart.LineChart
+import com.github.se.polyfit.ui.components.DropDownMenu
+import com.github.se.polyfit.ui.components.lineChartData
+import com.github.se.polyfit.ui.viewModel.GraphViewModel
+import com.github.se.polyfit.ui.viewModel.SortPoints
+
+@Composable
+fun FullGraphScreen() {
+  val isTestEnvironment = System.getProperty("isTestEnvironment") == "true"
```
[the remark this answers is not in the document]
sharif-bbd: I personally didn't find any library that would allow us to reach a satisfying visual result.

app/src/main/java/com/github/se/polyfit/ui/components/Graphs.kt
```diff
@@ -0,0 +1,81 @@
...
+import co.yml.charts.ui.linechart.model.LinePlotData
+import co.yml.charts.ui.linechart.model.LineStyle
+import co.yml.charts.ui.linechart.model.LineType
+import co.yml.charts.ui.linechart.model.SelectionHighlightPoint
+import co.yml.charts.ui.linechart.model.SelectionHighlightPopUp
+import co.yml.charts.ui.linechart.model.ShadowUnderLine
+import com.github.se.polyfit.ui.viewModel.DataToPoints
+
+@Composable
+@Preview
```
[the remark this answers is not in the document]
sharif-bbd: yes the preview doesn't appear here because this only contains the data for the composable not the actual composable. The composable is called in the GraphScreen

app/src/main/java/com/github/se/polyfit/ui/components/Graphs.kt
```diff
@@ -0,0 +1,81 @@
...
+fun lineChartData(): LineChartData {
+  val steps = 10
+  val points = DataToPoints()
```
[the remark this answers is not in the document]
sharif-bbd: doing it in 2 separate funcitons won't add anything as this is the data for the composable not the actual composable, so the preview won't show anything anyway so I'll take it out. I will however put the datapoints as parameters for the function

app/src/main/java/com/github/se/polyfit/ui/components/Graphs.kt
```diff
@@ -0,0 +1,81 @@
+package com.github.se.polyfit.ui.components
+
```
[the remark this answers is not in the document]
sharif-bbd: There isn't an Option to give name the axis on the graph but I could give a title and change the labels on the graduations to be more accurate

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
```diff
@@ -0,0 +1,99 @@
...
+        Row(
+            modifier = Modifier.fillMaxWidth().padding(5.dp),
+            horizontalArrangement = Arrangement.SpaceBetween) {
+              OutlinedTextField(
+                  value = searchText,
+                  onValueChange = viewModel::onSearchTextChanges,
+                  modifier = Modifier.fillMaxWidth(0.4f).testTag("GraphScreenSearchBar"),
+                  placeholder = { Text(text = "Search") },
+                  shape = MaterialTheme.shapes.large)
+              DropDownMenu(
```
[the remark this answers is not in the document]
sharif-bbd: I managed to find a fix but I don't know if it is fool proof

app/src/main/java/com/github/se/polyfit/ui/viewModel/GraphViewModel.kt
sharif-bbd: done !

app/src/main/java/com/github/se/polyfit/ui/viewModel/GraphViewModel.kt
```diff
@@ -0,0 +1,128 @@
...
+        GraphData(kCal = 1000.0, day = 15, month = "Mar.", weight = 35.0),
+        GraphData(kCal = 2399.3, day = 16, month = "Mar.", weight = 78.0),
+        GraphData(kCal = 2438.0, day = 17, month = "Mar.", weight = 80.2))
+
+fun DataToPoints(): List<Point> {
+  val data = mockData
+  return data.mapIndexed { index, graphData ->
+    Point(x = index.toFloat(), y = graphData.kCal.toFloat())
+  }
+}
```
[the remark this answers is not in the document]
sharif-bbd: this data cannot be previewed but I did set the mock data to private, although this part will be modified after the merge when I link the actual Data

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
sharif-bbd: Alright great idea!

app/src/main/java/com/github/se/polyfit/ui/components/Graphs.kt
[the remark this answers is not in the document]
sharif-bbd: I managed to do it, but the 1st date is cropped on the x axis and I didn't manage to find a way around as public libraries can be pretty tough to work with, but the functionnality is working. I created a Data Class for the date that might be removed or changed later on it is said in the comments

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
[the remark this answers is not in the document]
RandomUsername1315: Okay, that is not ideal, but lets just consider this technical debt that could solved in version 2 of the app.

app/src/main/java/com/github/se/polyfit/ui/components/Graphs.kt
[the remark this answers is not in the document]
RandomUsername1315: Okay, I see, but it might be a good idea to follow this pattern when you have @Preview to make the code more readable and facilitate integration with the class that will retrieve data for the graphs.

RandomUsername1315 (approved):
Awesome improvements and thank you for making such a cool ui for the graphs !

app/src/main/java/com/github/se/polyfit/ui/components/DropDown.kt
```diff
@@ -0,0 +1,64 @@
...
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DropDownMenu(
+    isExpanded: MutableState<Boolean>,
+    items: List<String>,
+    text: MutableState<String>,
+    onItemSelect: (String) -> Unit
+) {
+  val charSize = MaterialTheme.typography.titleMedium.fontSize.value.toInt()
+  var estimatedWidth by remember { mutableStateOf(100) }
```
medwardson, line 31: where does this value come from?

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
```diff
@@ -0,0 +1,126 @@
...
+import com.github.se.polyfit.ui.viewModel.DateList
+import com.github.se.polyfit.ui.viewModel.GraphViewModel
+import com.github.se.polyfit.ui.viewModel.SortPoints
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun FullGraphScreen(viewModel: GraphViewModel = hiltViewModel<GraphViewModel>()) {
+
+  val isTestEnvironment = System.getProperty("isTestEnvironment") == "true"
+  val isExpanded = remember { mutableStateOf(false) }
```
medwardson, line 47: what is the reason behind having this in this file rather than DropDown? It seems like it is only used and mutated there, and could be local to that file?

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
```diff
@@ -0,0 +1,126 @@
...
+
+  val sortedBy = remember { mutableStateOf("KCAL") }
+
+  val searchText by viewModel.searchText.observeAsState()
+  val graphData by viewModel.graphData.observeAsState()
+  // val isSearching by viewModel.isSearching.collectAsState()
```
medwardson, line 53:
```suggestion

```

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
```diff
@@ -0,0 +1,126 @@
...
+  Scaffold(
+      topBar = {
+        // This will be given a return function later
+        SimpleTopBar(LocalContext.current.getString(R.string.graphScreenTitle), {})
```
medwardson, line 57: Assuming the return function is coming from a navigation function passed, the composable header should include a `navigateBack` function that you use here, and then the default value can come from where this is called

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
```diff
@@ -0,0 +1,126 @@
...
+                  items(it) { data ->
+                    Spacer(modifier = Modifier.height(5.dp))
+                    Row(
+                        modifier = Modifier.fillMaxWidth().testTag("ElementsRow"),
+                        horizontalArrangement = Arrangement.SpaceBetween) {
+                          Column(
+                              horizontalAlignment = Alignment.Start,
+                              modifier = Modifier.padding(start = 4.dp)) {
+                                Text(
+                                    text = "${data.kCal} kCal", modifier = Modifier.testTag("kcal"))
```
medwardson, line 110: If we want to use values in `strings.xml`, how come we don't use it here? I believe you can pass arguments?

app/src/main/res/values/strings.xml
```diff
@@ -16,7 +16,11 @@
     <string name="condition3">" and "</string>
     <string name="condition4">"Privacy Policy"</string>
     <string name="signGoogle">"Sign in with Google"</string>
+
+    <string name="graphScreenTitle">Calories Per Day</string>
+
```
medwardson, line 21: extra space?

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
```diff
@@ -0,0 +1,126 @@
...
+                                Text(
+                                    text = "${data.weight} lbs",
+                                    modifier = Modifier.testTag("weight"))
+                              }
+                          Text(
+                              modifier = Modifier.padding(end = 4.dp).testTag("Date"),
+                              text = "${data.day} ${data.month}")
```
medwardson, line 117: same question about having string values stored external

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
```diff
@@ -0,0 +1,126 @@
...
+                    Row(
+                        modifier = Modifier.fillMaxWidth().testTag("ElementsRow"),
+                        horizontalArrangement = Arrangement.SpaceBetween) {
+                          Column(
+                              horizontalAlignment = Alignment.Start,
+                              modifier = Modifier.padding(start = 4.dp)) {
+                                Text(
+                                    text = "${data.kCal} kCal", modifier = Modifier.testTag("kcal"))
+                                Text(
+                                    text = "${data.weight} lbs",
```
medwardson, line 112: here too

app/src/main/java/com/github/se/polyfit/ui/components/scaffold/SimpleTopBar.kt
```diff
@@ -22,7 +23,7 @@ fun SimpleTopBar(title: String, navigateBack: () -> Unit) {
       title = {
         Text(
             title,
-            modifier = Modifier.testTag("${titleCase(title)} Title"),
+            modifier = Modifier.testTag("${titleCase(title)} Title").fillMaxWidth(),
```
medwardson, line 26: Why fill max width? If its to center the title, see https://github.com/swent-group10/polyfit/blob/b4137f6448ef50988c1a05e0bf290bd28a543b44/app/src/main/java/com/github/se/polyfit/ui/components/scaffold/CenteredTopBar.kt coming in #187 which I think gives a better solution, since we don't always want title in the center.

app/src/main/java/com/github/se/polyfit/ui/utils/GraphData.kt
```diff
@@ -0,0 +1,22 @@
+package com.github.se.polyfit.ui.utils
+
+data class GraphData(val kCal: Double, val day: Int, val month: String, val weight: Double) {
```
medwardson, line 3: it would be a good idea to write unit tests for this

app/src/main/java/com/github/se/polyfit/MainActivity.kt
```diff
@@ -42,6 +43,7 @@ class MainActivity : ComponentActivity() {
         val navController = rememberNavController()
         val navigation = Navigation(navController)
         NavHost(navController = navController, startDestination = Route.Register) {
+          composable(Route.Graph) { FullGraphScreen() }
```
medwardson, line 46: here is where we should pass in a back navigate to be used in the graph screen
```suggestion
          composable(Route.Graph) { FullGraphScreen(navigateBack = navigation::....) }
```

app/build.gradle.kts
```diff
@@ -208,6 +208,35 @@ android {
...
+        kapt("com.google.dagger:hilt-android-compiler:2.51")
+        kapt("com.google.dagger:hilt-compiler:2.51")
+        kaptAndroidTest("com.google.dagger:hilt-compiler:2.51")
+        kaptTest("com.google.dagger:hilt-compiler:2.51")
+        // Mockito
+        androidTestImplementation("org.mockito:mockito-core:5.11.0")
+
+        testImplementation("io.mockk:mockk:1.13.10")
+        androidTestImplementation("io.mockk:mockk:1.13.10")
+        androidTestImplementation("io.mockk:mockk-android:1.13.10")
```
medwardson, line 222: it seems like some of these already exist in the file above?

medwardson (changes requested):
Looks good, but a couple changes to keep consistent

app/src/main/java/com/github/se/polyfit/MainActivity.kt
sharif-bbd: Yes this will be done afterward I just wanted to finish this screen before linking it.

app/src/main/java/com/github/se/polyfit/ui/components/scaffold/SimpleTopBar.kt
sharif-bbd: Its because for some reason that when I didn't fill the max Width the title didn't display at all.

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
sharif-bbd: You're right

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
sharif-bbd: This will be implemented afterwards

app/src/main/res/values/strings.xml
sharif-bbd: this came from merging didn't pay attention

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
sharif-bbd: it seemed simpler as the data isn't fixed

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
sharif-bbd: the String.xml is usually intended for static strings which isn't the case here

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
sharif-bbd: same

app/src/main/java/com/github/se/polyfit/ui/utils/GraphData.kt
sharif-bbd: will do when I implement the Data as it will probably modified a little to accommodate

app/build.gradle.kts
sharif-bbd: you're right although they were already duplicates previous to me adding them

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
medwardson: Even though the data is not fixed it still follows a format, and then you can use the strings.xml and pass a variable in. I think its better to either use this for everything or nothing at all, since it doesn't make sense to use it conditionally

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
medwardson: I don't think it is unique for static strings: https://developer.android.com/develop/ui/compose/resources#strings

app/src/main/java/com/github/se/polyfit/ui/components/scaffold/SimpleTopBar.kt
medwardson: That seems unexpected, but I don't think this is the right solution, as this component is already working in other screens

app/build.gradle.kts
medwardson: Gotcha, eitherway its best to remove adding them more to prevent clutter

app/src/androidTest/java/com/github/se/polyfit/ui/screen/GraphScreenTest.kt
```diff
@@ -0,0 +1,298 @@
...
+  @get:Rule val composeTestRule = createComposeRule()
+  @get:Rule val mockkRule = MockKRule(this)
+
+  @Before
+  fun setup() {
+    System.setProperty("isTestEnvironment", "true")
+    mockkStatic(Log::class)
+    /*
+
+    */
```
medwardson, line 49:
```suggestion
```

app/src/main/res/values/strings.xml
```diff
@@ -15,7 +15,12 @@
     <string name="condition3">" and "</string>
     <string name="condition4">"Privacy Policy"</string>
     <string name="signGoogle">"Sign in with Google"</string>
+    <string name="graphScreenTitle">Calories Per Day</string>
+    <string name="kcalvalue">%1$.1f kCal</string>
+    <string name= "weightvalue">%1$.1f lbs</string>
```
medwardson, line 20:
```suggestion
    <string name="weightvalue">%1$.1f lbs</string>
```

medwardson (approved):
Looks great, thanks for taking the time with the review changes 🙏

PR #179, opened 2024-04-28
Daily Recap Screen

medwardson:
Screen to give recaps of meals from the selected day.
- Defaults to current day and displays all meals
- Can switch to past days and will fetch updated meals
- After creating/updating a meal, it clears the ViewModel

TODO:
- Make necessary changes to AddMealFlow to allow editing meal, as currently it is not set up properly.


app/src/main/java/com/github/se/polyfit/ui/screen/DailyRecapScreen.kt
```diff
@@ -0,0 +1,63 @@
...
+  val occasions = MealOccasion.entries.filter { it != MealOccasion.OTHER }
+
+  Scaffold(topBar = { SimpleTopBar("Overview", navigateBack) }) {
+    Column(modifier = Modifier.padding(it).testTag("DailyRecapScreen")) {
+      DateSelector(onConfirm = dailyRecapViewModel::setDate, title = "Date")
+      if (isFetching) {
+        CircularProgressIndicator(
+            modifier = Modifier.fillMaxSize().padding(16.dp).testTag("Spinner"))
+      }
+      if (!isFetching) {
```
medwardson, line 42: Chose to do this instead of if/else as it seems more clear

app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
```diff
@@ -94,8 +100,8 @@ private fun BottomBar(setMeal: () -> Unit, isComplete: Boolean, navigateForward:
               PrimaryButton(
                   onClick = {
                     Log.v("Add to Diary", "Clicked")
-                    setMeal()
                     navigateForward()
+                    setMeal()
```
medwardson, line 104: Prevent UI from updating before navigating away

app/src/main/java/com/github/se/polyfit/ui/components/list/MealList.kt
```diff
@@ -0,0 +1,98 @@
...
+            style = MaterialTheme.typography.bodyLarge,
+            color = PurpleGrey40,
+            modifier = Modifier.testTag("MealName"))
+        Text(
+            "${meal.calculateTotalCalories()} ${MeasurementUnit.KCAL.name.lowercase()}",
+            style = MaterialTheme.typography.bodyMedium,
+            color = PurpleGrey40,
+            modifier = Modifier.testTag("MealCalories"))
+      }
+      MealTagList(
```
aprilslis, line 90: Very good decision on reusing the components!

app/src/main/java/com/github/se/polyfit/ui/screen/DailyRecapScreen.kt
```diff
@@ -0,0 +1,63 @@
...
+          }
+        }
+      }
+    }
+  }
+}
+
+@Preview
+@Composable
+fun DailyRecapScreenPreview() {
```
aprilslis, line 61: Do we need to comment this out or remove this as it only for development process?

app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
```diff
@@ -39,13 +39,19 @@ fun NutritionScreen(
     navigateForward: () -> Unit
 ) {
   val isComplete by mealViewModel.isComplete.collectAsState()
+
+  fun setMeal() {
```
aprilslis, line 43: Is there any reason why this bug fix is included in this pr?

app/src/main/java/com/github/se/polyfit/ui/theme/Color.kt
```diff
@@ -16,3 +16,4 @@ val GradientButtonPink = Color(0xFF4A00E0)
 val GradientButtonPurple = Color(0xFF8E2DE2)
 val GradientButtonGrey = Color(0xFFE0E0E0)
 val DeleteIconRed = Color(0xFFFF7575)
+val BorderPurple = Color(0xFFD8D1F5)
```
aprilslis, line 19: No action needed for this PR, but we should add a ticket to backlog to let the overview screen use this proper border color as well

aprilslis (approved):
Looks pretty good overall! Just some small questions but other than that lgtm!

app/src/main/java/com/github/se/polyfit/ui/screen/DailyRecapScreen.kt
medwardson: I don't think it needs to be commented out, but I can do it if that's preferred. It shouldn't interrupt anything

app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
medwardson: Yes, this is a set of integrating the ability to edit meals, which was worked on in this PR but is delayed since the AddMealFlow is not fully ready. This is still an important change to include

app/src/main/java/com/github/se/polyfit/viewmodel/dailyRecap/DailyRecapViewModel.kt
```diff
@@ -0,0 +1,41 @@
...
+    ViewModel() {
+  private val _meals: MutableStateFlow<List<Meal>> = MutableStateFlow(mutableListOf())
+  val meals: StateFlow<List<Meal>> = _meals
+
+  private val _isFetching: MutableStateFlow<Boolean> = MutableStateFlow(false)
+  val isFetching: StateFlow<Boolean> = _isFetching
+
+  init {
+    getMealsOnDate(LocalDate.now())
+  }
```
sharif-bbd, line 27: Be careful not with this line, if the viewmodel gets instantiated frequently it will result in multiple unnecessary fetching of the data. If this happens currently or later on an Idea would be to put a refresh mechanism on it and start fetching only when the user refreshes

sharif-bbd (commented):
Really good PR, I would suggest handling the case where the occasions list is empty as I feel showing a blank screen isn't really good practice. But nevertheless Excellet Job !

app/src/main/java/com/github/se/polyfit/ui/components/list/MealList.kt
```diff
@@ -0,0 +1,98 @@
...
+    mealViewModel: MealViewModel = hiltViewModel()
+) {
+  val filteredMeals = meals.filter { it.occasion == occasion }
+  val totalCalories = filteredMeals.sumOf { it.calculateTotalCalories() }.toString()
+
+  Surface(
+      modifier = modifier.fillMaxWidth().padding(8.dp, 0.dp).testTag("MealList"),
+      border = BorderStroke(1.dp, BorderPurple),
+      shape = RoundedCornerShape(4.dp),
+      color = Color.White) {
```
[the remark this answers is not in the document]
medwardson: I don't think our material theme matches the figma design. Maybe this can be a future ticket, but for now this and many other parts of the code base don't follow because it's not up to date with Figma's colors

app/src/main/java/com/github/se/polyfit/ui/components/list/MealList.kt
```diff
@@ -0,0 +1,98 @@
...
+
+@Composable
+fun MealList(
+    meals: List<Meal>,
+    occasion: MealOccasion,
+    navigateTo: () -> Unit,
+    modifier: Modifier = Modifier,
+    mealViewModel: MealViewModel = hiltViewModel()
+) {
+  val filteredMeals = meals.filter { it.occasion == occasion }
```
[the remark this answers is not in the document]
medwardson: It is intended for no meals to show up for a specific occasion. I think the comment you added below for if there is no meals anywhere suits well, but it's ok to show an empty list if one occasion is empty

app/src/main/java/com/github/se/polyfit/viewmodel/dailyRecap/DailyRecapViewModel.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.coroutines.withContext
+
+@HiltViewModel
+class DailyRecapViewModel @Inject constructor(private val mealRepository: MealRepository) :
+    ViewModel() {
+  private val _meals: MutableStateFlow<List<Meal>> = MutableStateFlow(mutableListOf())
+  val meals: StateFlow<List<Meal>> = _meals
+
+  private val _isFetching: MutableStateFlow<Boolean> = MutableStateFlow(false)
+  val isFetching: StateFlow<Boolean> = _isFetching
```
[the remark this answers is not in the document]
medwardson: StateFlow also reacts to events that update the values (i.e. if you switch to an old date that has meals, when we set the meal value it will refresh the screen. Further, it provides default values versus having potential null values, which is why I went with it. Do you think that LiveData will provide any additional benefits on this purpose?

app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
medwardson: This is not included in the ViewModel because the ViewModel should not represent multiple instances. This is a temporary solution until https://github.com/swent-group10/polyfit/issues/164 is finished

RandomUsername1315: I am working on it. Except a new implementation by tomorrow.

PR #187, opened 2024-04-30
UI/ Helpers for CreatePostScreen

aprilslis:
These are the helper components for UI of making a new post for others to see. (see #158 for final look of screen.)

Components:
- Picture selector for picking a picture for the post (only a place holder now as we do not yet know if backend will be able to support image for posts)
- Textfield for post description input
- Meal selector for picking a meal for the post
- Meal Details for displaying details of the meal selected for the post
- Alert Dialog for asking for use permission to use their location

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/model/meal/Meal.kt
+++ b/app/src/main/java/com/github/se/polyfit/model/meal/Meal.kt
@@ -2,6 +2,7 @@ package com.github.se.polyfit.model.meal
 
 import android.util.Log
 import com.github.se.polyfit.model.ingredient.Ingredient
+import com.github.se.polyfit.model.nutritionalInformation.Nutrient
 import com.github.se.polyfit.model.nutritionalInformation.NutritionalInformation
 import java.time.LocalDate
 
@@ -107,6 +108,21 @@ data class Meal(
     return name.isNotEmpty() &&
         ingredients.isNotEmpty() &&
         nutritionalInformation.nutrients.isNotEmpty()
+  }
+
+  fun getNutrient(nutrientType: String): Nutrient? {
+    return this.nutritionalInformation.getNutrient(nutrientType)
+  }
+
+  fun getMacros(): Map<String, Double> {
+    val macros = mutableMapOf<String, Double>()
+    val protein = getNutrient("protein")?.amount ?: 0.0
+    val fat = getNutrient("fat")?.amount ?: 0.0
+    val carbs = getNutrient("carbohydrates")?.amount ?: 0.0
+    macros["protein"] = protein
+    macros["fat"] = fat
+    macros["carbohydrates"] = carbs
+    return macros
   }
 
   private fun updateMeal() {
--- a/app/src/main/java/com/github/se/polyfit/model/post/Post.kt
+++ b/app/src/main/java/com/github/se/polyfit/model/post/Post.kt
@@ -28,15 +28,15 @@ data class Post(
   }
 
   fun getCarbs(): Nutrient? {
-    return meal.nutritionalInformation.getNutrient("carb")
+    return meal.getNutrient("carbohydrates")
   }
 
   fun getFat(): Nutrient? {
-    return meal.nutritionalInformation.getNutrient("fat")
+    return meal.getNutrient("fat")
   }
 
   fun getProtein(): Nutrient? {
-    return meal.nutritionalInformation.getNutrient("protein")
+    return meal.getNutrient("protein")
   }
 
   fun serialize(): Map<String, Any> {
--- a/app/src/main/java/com/github/se/polyfit/ui/components/dialog/LocationPermissionDialog.kt
+++ b/app/src/main/java/com/github/se/polyfit/ui/components/dialog/LocationPermissionDialog.kt
@@ -1,0 +1,64 @@
+package com.github.se.polyfit.ui.components.dialog
+
+import android.content.Context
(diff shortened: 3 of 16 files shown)


app/src/main/java/com/github/se/polyfit/model/meal/Meal.kt
```diff
@@ -109,6 +110,21 @@ data class Meal(
         nutritionalInformation.nutrients.isNotEmpty()
   }
 
+  fun getNutrient(nutrientType: String): Nutrient? {
+    return this.nutritionalInformation.getNutrient(nutrientType)
+  }
+
+  fun getMacros(): Map<String, Double> {
```
medwardson, line 117: Good idea to group it together

app/src/main/java/com/github/se/polyfit/ui/components/dialog/LocationPermissionDialog.kt
medwardson, line 1: I'm not sure if this is needed or not, as I think android might provide a default popup to request location permission. I think its ok to leave it for now however, since, location isn't merged into main yet

app/src/main/java/com/github/se/polyfit/ui/components/scaffold/CenteredTopBar.kt
medwardson, line 1: Thanks for making it a separate component for future use 👍

app/src/main/java/com/github/se/polyfit/ui/components/textField/MealInputTextField.kt
```diff
@@ -0,0 +1,101 @@
...
+    onCollapseSearchBar: () -> Unit = {},
+    onMealChange: (Meal) -> Unit = {}
+) {
+  val context = LocalContext.current
+  var showIngredientSearch by remember { mutableStateOf(false) }
+  var searchText by remember { mutableStateOf("") }
+  var searchResults by remember { mutableStateOf(meals) }
+
+  // Set the search bar to be at most 5 items tall, then scroll
+  val maxHeight = (74 + minOf(5, searchResults.size) * 41).dp
```
medwardson, line 47: Usually, it is best to find a more responsive way to do height, without hard calculations. I think this is ok for now, but after M2 maybe we can revisit this

app/src/main/java/com/github/se/polyfit/ui/components/textField/MealInputTextField.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  SearchBar(
```
medwardson, line 49: Is there a reason we chose a search bar instead of dropdown menu here?

app/src/main/java/com/github/se/polyfit/ui/components/selector/MealSelector.kt
```diff
@@ -0,0 +1,162 @@
...
+                Pair("Carbs", macros["carbohydrates"]),
+                Pair("Fat", macros["fat"]),
+                Pair("Protein", macros["protein"]),
+            )
+
+        nutrients.forEach { (nutrientName, amount) ->
+          GradientBox(
+              outerModifier = Modifier.padding(horizontal = 10.dp),
+              innerModifier =
+                  Modifier.heightIn(min = 80.dp, max = 80.dp).widthIn(max = 80.dp, min = 80.dp),
```
medwardson, line 147: if the range is not actually a difference, why not just use 
```suggestion
                  Modifier.height(80.dp).width(80.dp),
```

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
```diff
@@ -26,15 +26,15 @@ data class Post(
   }
 
   fun getCarbs(): Nutrient? {
-    return meal.nutritionalInformation.getNutrient("carb")
+    return meal.getNutrient("carbohydrates")
```
medwardson, line 29: Good catch

medwardson (approved):
Left a few questions, but nothing major. Nice job

app/src/main/java/com/github/se/polyfit/ui/components/selector/MealSelector.kt
```diff
@@ -0,0 +1,162 @@
...
+              modifier = Modifier.fillMaxWidth().padding(4.dp).testTag("Ingredient"),
+              verticalAlignment = Alignment.CenterVertically) {
+                Text(
+                    text = ingredient.name,
+                    modifier = Modifier.weight(1f).padding(start = 10.dp),
+                    fontSize = 16.sp,
+                    color = SecondaryGrey)
+
+                Text(
+                    text = "${ingredient.amount} ${ingredient.unit.toString().lowercase()}",
```
medwardson, line 122: You can use resource strings for this too, and pass in params

app/src/main/java/com/github/se/polyfit/ui/components/textField/MealInputTextField.kt
aprilslis: Ok. For now, it wasn't resizing well on its own and this was the best way

app/src/main/java/com/github/se/polyfit/ui/components/selector/MealSelector.kt
aprilslis: I think its good for static resource but this is just two variable together, so it will be ok

PR #191, opened 2024-04-30
Issue with the "Add to Diary" not being enable when it should

RandomUsername1315:
## Bug Report: "Add to Diary" Button Not Being Enabled When It Should

**Issue Summary:**  
Issue with how the "Add to Diary" button is enabled on the last screen when creating a meal. Sometimes, even if all the right data are entered correctly, the button stays disabled.

**Steps to Reproduce:**
1. Open the app and go to the overview page.
2. Create a meal (either through image recognition or manual input).
3. Click on the option to create a meal manually and then go back to the main screen.
4. Follow the steps to create a meal.
5. At the end, the "Add to Diary" button is disabled.

**Expected Result:**  
The button should be enabled.

**Video Proof:**  
[Video Link](https://github.com/swent-group10/polyfit/assets/100274930/9d868f37-c8de-4852-8b19-54b367cd913a)

---

### Issue Details
The issue is with the following line of code in `MealViewModel.kt` on line 29:
```kotlin
init {
    viewModelScope.launch { _meal.collect { meal -> _isComplete.value = meal.isComplete() } }
}
```

Essentially, this sets a listener on _meal to update `_isComplete`. The problem is that this is launched within the ViewModel's scope. When the ViewModel is no longer in use, for example, by going back to the overview screen, it seems that all coroutines in that scope are terminated, which results in the listener being killed. 

## Proposed Temporary Fix
To fix this, the new implementation uses the global scope.

## Long-Term Improvement
For a more permanent solution, creating a new ViewModel for each meal creation instead of reusing the same one. This approach avoids issues related to lifecycle management of the ViewModel's coroutines.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -162,12 +162,8 @@ android {
         kaptAndroidTest("com.google.dagger:hilt-compiler:2.51")
         kapt("androidx.hilt:hilt-compiler:1.2.0")
 
-        // Hilt
-        implementation("androidx.hilt:hilt-navigation-compose:1.2.0")
-
         // Android Test
         androidTestImplementation("com.squareup.okhttp3:okhttp:4.12.0")
-        androidTestImplementation("androidx.arch.core:core-testing:2.2.0")
         androidTestImplementation("androidx.test.espresso:espresso-core:3.5.1")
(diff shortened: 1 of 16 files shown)


medwardson (approved):
Thanks for picking this ticket up, it works as expected on my end.

PR #192, opened 2024-05-01
meal view model creation

RandomUsername1315:
# Pull Request Overview

This pull request focuses on addressing technical debt introduced in Milestone 1. Specifically, it implements a new `MealViewModel` for each iteration through the flow of adding a meal. This approach is beneficial for several reasons. Firstly, it prevents cross-contamination between meals. For instance, users previously encountered a bug where, after creating a meal, the ingredients from the last meal would reappear when they attempted to create a new one. Using separate view models for each meal ensures that each ingredient list starts empty. Additionally, this change narrows the scope of "observers" to the ViewModel level, enhancing the encapsulation and reliability of the data flow.

## Changes
Numerous incremental changes were made, stemming from minor adjustments that enabled more significant architectural improvements, including:

- **Create `OverviewViewModel.kt` Class**: A new class designed to manage all logic related to storing a new meal from a picture, ensuring a clear separation between logic and UI.
  
- **Create `Post.kt` Class**: Introduces a data structure that encapsulates the information of a post.

- **Modify the `Route.AddMeal`**: Updated to accept the new meal's ID as an argument, facilitating the transition from meal overview to the add meal flow within the app.

## Implementation Checklist

- [x] OverviewModel
- [x] Fix the meal dao to return a valid id for a stored meal

## Tests Checklist

- [x] Fix the broken tests.
- [x] Test OverviewModel.
- [x] Expand testing to increase coverage.

## Questions for Reviewers

The current setup is such that if a user creates a meal via image recognition, it is automatically added to the database. If the user opts to return to the main menu without completing the meal creation process, the meal will still persist in the database. How do we want to handle uncomplete meals? Delete them ?


RandomUsername1315:
There seems to be an issue with Github. This code was forked from the main branch, but github does not seem to recognize code that was already present on the main branch is it counting it as new code. I will have to do some digging to resolve this.

app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
```diff
@@ -27,14 +27,26 @@ class MealViewModel @Inject constructor(private val mealRepo: MealRepository) :
...
-      _meal.map { it.isComplete() }.stateIn(GlobalScope, SharingStarted.Eagerly, false)
+      _meal.map { it.isComplete() }.stateIn(viewModelScope, SharingStarted.Eagerly, false)
   val isComplete: StateFlow<Boolean>
     get() = _isComplete
 
   fun setMealData(meal: Meal) {
     _meal.value = meal
   }
 
+  fun setMealData(mealId: Long) {
```
medwardson, line 38: What are your thoughts of making this part of the `MealViewModel` init flow? If we are no longer reusing viewmodels, we will never need this function other than when the ViewModel is created. So, we can make an optional parameter in the constructor for mealId, and then run this there

medwardson (commented):
It looks good! I'd suggest making the `MealViewModel` accept the mealId as a constructor param and remove the setMeal/reset methods, since it should be specific to each instance. My other suggestion is to extract some of the functions out of the OverviewViewModel, as I believe they could be relevant to other parts of the code (thinking graphs) in the future.

As for your comment about deleting meals, we discussed offline, but, the gist of the result was that since the new meals created don't have a valid MealOccassion, they can be filtered out by that.

As for functionality, things work as expected. Nice job

app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
```diff
@@ -0,0 +1,168 @@
...
+    // Log the result
+    if (filteredMeals.isEmpty()) {
+      Log.e("OverviewViewModel", "No meals found in the specified calorie range")
+    } else {
+      Log.i("OverviewViewModel", "Found ${filteredMeals.size} meals in the specified calorie range")
+    }
+
+    // Return the filtered meals
+    return filteredMeals
+  }
```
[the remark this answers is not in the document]
RandomUsername1315: I know in reality it should be. But the modifications I made for the meal view model, are in places extremely hard to tests such as the main activity. So I had to pad this pr with this to reach the required 80%. I know it is bad practice, but I could not find any other way.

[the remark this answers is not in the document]
medwardson: I understand - if I was allowed to force approve it separately I would, but this makes sense then!

app/src/main/java/com/github/se/polyfit/viewmodel/meal/MealViewModel.kt
RandomUsername1315: That's a great idea. I am a bit hesitant to implement that because all the arguments are injected by Hilt. I'll give it a try to see if I can make it work.
Update : 
The issue is that because the ViewModel is created by Hilt, we cannot dynamically pass arguments to it when it's created. So, I don't think this is possible. I did clean up the implementation a bit by changing the argument of the function from Long to Long? and removing the if condition in the `AddMealFlow`.

app/src/main/java/com/github/se/polyfit/ui/screen/IngredientScreen.kt
```diff
@@ -43,7 +43,6 @@ fun IngredientScreen(
 
   fun goBackAndReset() {
     navigateBack()
```
medwardson, line 45:
```suggestion
```

app/src/main/java/com/github/se/polyfit/ui/screen/IngredientScreen.kt
```diff
@@ -43,7 +43,6 @@ fun IngredientScreen(
...
-    mealViewModel.reset()
   }
 
   Scaffold(
```
medwardson, line 48:
```suggestion
  Scaffold(
    topBar = { TopBar({ navigateBack }) }
```

can just remove the goBackAndReset fn

medwardson (approved):
Looks great, good job

PR #194, opened 2024-05-01
Small fix to the sign-in processus 

Neb1010:
## Before
When sign-in the users were not correctly sign-up to the firebase project.
I suppose the previous code was just checking if it was a real google account.
## After
The user is now sign-up to the firebase and has a "token" that is necessary to fetch data from the storage


RandomUsername1315 (approved):
Thank you for finding this bug ! Amazing job !!!

PR #195, opened 2024-05-01
Fix: Nutrient name formatting

medwardson:
The response from the API body is in the format `Some Food Name` rather than `someFoodName` as the original API had intended. Because of this, there is errors in formatting things like vitamins. I did a clean up of the codebase to make sure that we are using the proper formatting so that the vitamin names format as expected.


app/src/androidTest/java/com/github/se/polyfit/ui/screen/IngredientTest.kt
```diff
@@ -201,13 +201,6 @@ class IngredientTest : TestCase() {
         performClick()
       }
 
-      val ingredients =
```
medwardson, line 204: dead code

antomax33 (approved):
LGTM

PR #201, opened 2024-05-02
Ignore case for nutrient searching

medwardson:
Nutrient searching should not care about case sensitivity.


RandomUsername1315 (approved):
Nice catch ! Thank you for cleaning up the code and making it more consistent !

Issue #159, opened 2024-04-25
GeoFire


Neb1010:
Implementation of [GeoFire](https://firebaseopensource.com/projects/firebase/geofire-android/)

PR #204, opened 2024-05-02
Adds the backends to view the list of posts

RandomUsername1315:
# Pull Request Overview

This PR establishes the links between the backend for posts and the interface for scrolling through posts. Due to time constraints, only the UI that lists the posts is implemented.

## Changes

- **Create `ViewPostViewModelTest.kt` Class**: A simple class that handles basic calls to fetch the posts.
- **Modify `MainActivity.kt` Class**: Modified to allow navigation to the new UI.
- **Modify the `PostFirebaseRepository`**: Updated to fetch posts from Firebase.

## Implementation Checklist

- [x] ViewPostViewModelTest
- [x] MainActivity
- [x] PostFirebaseRepository

## Tests Checklist

- [x] Fix the broken tests.
- [x] Live test with PostFirebaseRepository
- [x] Expand testing to increase coverage.

## Future PRs
- Add the map UI.
- Add the ability to store images associated with posts.

## Acknowledgement

- The PR would not have been possible if not for @Neb1010 's code from the #159


app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
```diff
@@ -18,35 +15,25 @@ class PostFirebaseRepository(
...
-    try {
-      val uploadTask = refSource.putStream(stream)
-      return uploadTask.await().storage
-    } catch (e: Exception) {
-      Log.e("PostFirebaseRepository", "Failed to upload image", e)
-      throw Exception("Error uploading images : ${e.message}", e)
+    postCollection.get().await().map { document ->
+      Log.d("PostFirebaseRepository", "Document: ${document.data}")
+      Post.deserialize(document.data)?.let { posts.add(it) }
     }
```
antomax33, line 35: Try and cache, it can fail

app/src/main/java/com/github/se/polyfit/model/post/Location.kt
```diff
@@ -14,5 +14,13 @@ data class Location(
     fun default(): Location {
       return Location(0.0, 0.0, 0.0, "EPFL")
     }
+
+    fun deserialize(data: Map<String, Any>): Location {
+      return Location(
+          data["longitude"] as Double,
+          data["latitude"] as Double,
+          data["altitude"] as? Double ?: 0.0,
```
antomax33, line 22: Where do we need the altitude in the app?

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
```diff
@@ -70,6 +71,33 @@ data class Post(
...
+      } catch (e: Exception) {
+        Log.e("Post", "Failed to deserialize Post object: ${e.message}", e)
+        throw IllegalArgumentException("Failed to deserialize Post object", e)
+      }
+    }
+
+    private fun deserializeLocalDate(data: Map<String, Any?>, key: String): LocalDate? {
+      return try {
+        val dateString = data[key] as? String
+        dateString?.let { LocalDate.parse(it) }
```
antomax33, line 95: It is better to fail and go to the catch then return a null value.

app/src/main/java/com/github/se/polyfit/ui/screen/PostInfoScreen.kt
```diff
@@ -2,34 +2,53 @@ package com.github.se.polyfit.ui.screen
...
+import androidx.hilt.navigation.compose.hiltViewModel
 import com.github.se.polyfit.R
 import com.github.se.polyfit.model.post.Post
 import com.github.se.polyfit.ui.components.postinfo.PostCard
+import com.github.se.polyfit.viewmodel.post.ViewPostViewModel
 
 @Composable
-fun PostInfoScreen(posts: List<Post>, index: Int = 0) {
+fun PostInfoScreen(
+    posts: List<Post> = listOf(),
```
antomax33, line 30: It could be better to remove the posts and only use the ViewPostViewModel.

app/src/main/java/com/github/se/polyfit/ui/screen/PostInfoScreen.kt
```diff
@@ -2,34 +2,53 @@ package com.github.se.polyfit.ui.screen
...
+  val isFetching by viewPostViewModel.isFetching.collectAsState()
 
-  if (posts.isEmpty()) {
-    NoPost()
-    return
-  }
+  if (isFetching) {
+    Box(modifier = Modifier.fillMaxSize(), contentAlignment = Alignment.Center) {
+      CircularProgressIndicator(modifier = Modifier.padding(16.dp))
+    }
```
antomax33, line 40: We can directly return instead of the else part to have less nesting.

antomax33 (commented):
Still some smalls changes to do.

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
RandomUsername1315: I completely missed that thank you for catching it.

app/src/main/java/com/github/se/polyfit/ui/screen/PostInfoScreen.kt
RandomUsername1315: I didn't think that was possible in `Composable`. Thank you for the quick tip.

app/src/androidTest/java/com/github/se/polyfit/viewmodel/post/ViewPostViewModelTest.kt
medwardson, line 1: this name 😆

app/src/main/java/com/github/se/polyfit/model/post/Location.kt
RandomUsername1315: It is artifact of a previous version used to avoid breaking anything when fetching data from the firebase.

app/src/main/java/com/github/se/polyfit/ui/screen/PostInfoScreen.kt
RandomUsername1315: This is there for backwards compability to avoid breaking any previous implementation

antomax33 (approved):
LGTM

PR #205, opened 2024-05-02
Temporary access to create post

aprilslis:
Since we don't have designs for this yet, putting a temporary way to get to the CreatePostScreen that we can change after M2.


RandomUsername1315 (approved):
Looks good !

PR #206, opened 2024-05-02
Feature/backend/geofirewithpicture

Neb1010:
## Description

This PR introduces several updates to our Firebase integration for the `PostFirebaseRepository`, along with related models `Post` and `Location`. The updates aim to enhance the functionality and testability of our geo-querying capabilities and post management in the Firebase context.

## Changes

- `PostFirebaseRepository` : add function to fetch the right `Post` with they current images references in a certain radius, in a async way
- ~~.`PostFirebaseRepository` : change share function to push a node in the rtdb~~. has been removed on #203 he will handle this later 
- ~~.`Location` : deserialize a location~~. has been pushed on #203 
- ~~.`Post` : deserialize a Post from db and deserialize a `LocalDate`~~. has been pushed on #203

## How Has This Been Tested?

- Real test with the different database, and fetch only the post in a certain radius of a given location, with the images. references 
- Unit tests with MockK.

## Checklist

- [x] Get all posts in the data base within a certain radius.
- [x] Get the images references of the related Post to display them later.
- [x] Manage to pass some tests not related to my work

### Problem with the test
```
java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.IllegalStateException: 
Resource Compose-Espresso link isIdleNow() is returning true, but a message indicating 
that the resource has transitioned from busy to idle was never sent.
 at androidx.test.espresso.Espresso.onIdle(Espresso.java:18)
```
Did not directly managed to fix this error, even if all tests pass locally, some not related to the pr were failing on the CI.
but it's probably related to async tests, I read some people overpass this issues with an override of the espresso idle class.


app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
```diff
@@ -1,19 +1,39 @@
...
 import kotlinx.coroutines.tasks.await
+import kotlinx.coroutines.withContext
 
 class PostFirebaseRepository(
     private val db: FirebaseFirestore = FirebaseFirestore.getInstance(),
-    private val pictureDb: FirebaseStorage = FirebaseStorage.getInstance()
+    private val pictureDb: FirebaseStorage = FirebaseStorage.getInstance(),
+    private val rtdb: FirebaseDatabase =
+        FirebaseDatabase.getInstance(
+            "https://polyfit-316e8-default-rtdb.europe-west1.firebasedatabase.app/")
```
Neb1010, line 32: I know it's not a good practice, but it becomes really tricky to tests, I'll find a better way later

app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepositoryFetchPosts.kt
```diff
@@ -0,0 +1,98 @@
+/*
```
Neb1010, line 1: The tests that break other test on the CI.... Don't ask me why....

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
```diff
@@ -20,13 +21,9 @@ data class Post(
     override var location: Location,
     override var meal: Meal,
     override var createdAt: LocalDate,
-    var listOfImages: List<Bitmap> = emptyList()
+    var listOfImages: List<Bitmap> = emptyList(),
+    var listOfURLs: List<StorageReference> = emptyList()
```
RandomUsername1315, line 25: Why is storing both the image (the bitmap) and the storage reference needed? Is this for a future implementation or for backwards compatibility?

Also, it might be a better idea to use val and mutable lists instead of var. This would make adding new elements to the lists easier, as there would not be a need to create a new list each time.

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
RandomUsername1315: We can add that the pile of technical dept that we currently have (lol).
I had to do something very similar with Spoonacular API. I had to have the CI inject the API keys in the gradle.properties. Feel free to hit me up if you have any questions!

app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepositoryFetchPosts.kt
RandomUsername1315: Maybe something to do with asyc coroutines. 

Do you think this is something worth fixing in the future (making sure those tests pass)? If not, it might be a good idea to just delete this file to have a cleaner code base.

app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/GeoQueryRepositoryTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    every { mockGeoFireRef.equalTo("someStringValue" as String?) } returns mockGeoFireRef
+    every { mockGeoFireRef.addChildEventListener(any()) } returns mockk()
+
+    every { mockGeoFireRef.addListenerForSingleValueEvent(any()) } just Runs
+
+    repository = PostFirebaseRepository(rtdb = mockRtdb)
+  }
+
+  @Test
+  fun queryNearbyPostsShouldTriggerGeoQueryEventListener() = runBlocking {
```
RandomUsername1315, line 46: Since runBlocking is an older function that has been replaced by runTest, it might be a good idea to use the latter in the future.

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
```diff
@@ -1,19 +1,39 @@
...
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.tasks.await
+import kotlinx.coroutines.withContext
 
 class PostFirebaseRepository(
     private val db: FirebaseFirestore = FirebaseFirestore.getInstance(),
-    private val pictureDb: FirebaseStorage = FirebaseStorage.getInstance()
+    private val pictureDb: FirebaseStorage = FirebaseStorage.getInstance(),
```
RandomUsername1315, line 29: To be consistent with usage of Hilt in the rest of the app, it might be a smart idea to use dependency injection for both of the Firebase. The provides function should be added to the `UserModule.kt` file.

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
```diff
@@ -41,4 +61,110 @@ class PostFirebaseRepository(
...
+            }
+          }
+          .addOnFailureListener {
+            Log.e("PostFirebaseRepository", "Failed to fetch posts: ${it.message}")
+          }
+    }
+  }
+
+  suspend fun fetchImageReferencesForPost(postKey: String): List<StorageReference> {
+    return withContext(Dispatchers.IO) {
```
RandomUsername1315, line 164: In general, it is better to avoid using hardcoded Dispatchers.Default and instead allow Kotlin to optimize in the background which dispatcher to use.

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
```diff
@@ -41,4 +61,110 @@ class PostFirebaseRepository(
...
+      val storageRef = pictureDb.getReference("posts/$postKey")
```
RandomUsername1315, line 165: Maybe 'posts/' could be stored in a string resource file to avoid hardcoded references.

RandomUsername1315 (approved):
Looks good ! Just a few questions.

app/src/androidTest/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepositoryFetchPosts.kt
Neb1010: Yes, but I'd like too keep them at least until it's linked with the Map, after I'll remove this evil tests

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
Neb1010: listOfImages where the list of the pictures the user wants to add to the post, first I tried to fetch locally the pictures, it's doable but nobody in the industry do it like this but keep the picture references and display them whit Glide.

if we would have a cleaner code we should need to review and rewrite a big chunk of the code

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
Neb1010: changed by Default

Neb1010:
My pain is over until next pr

PR #217, opened 2024-05-03
fix not show every nutrient but only 3 principals

antomax33:
postinfo_title

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/ui/components/postinfo/Text.kt
+++ b/app/src/main/java/com/github/se/polyfit/ui/components/postinfo/Text.kt
@@ -22,16 +22,18 @@ import com.github.se.polyfit.model.post.Post
 fun TextPostInfo(modifier: Modifier = Modifier, post: Post) {
 
   Row(horizontalArrangement = Arrangement.SpaceBetween, modifier = Modifier.fillMaxWidth()) {
-    TextPost(
-        text = ContextCompat.getString(LocalContext.current, R.string.description),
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/se/polyfit/ui/components/postinfo/Text.kt
```diff
@@ -22,25 +22,29 @@ import com.github.se.polyfit.model.post.Post
...
 
   TextPost(
       text = ContextCompat.getString(LocalContext.current, R.string.nutrient),
       modifier = modifier.testTag("NutrientTitle"),
       bold = true)
 
-  post.meal.ingredients.forEach { it.nutritionalInformation.nutrients.forEach { NutrientInfo(it) } }
+  post.getCarbs()?.let { NutrientInfo(it) }
+  post.getFat()?.let { NutrientInfo(it) }
+  post.getProtein()?.let { NutrientInfo(it) }
```
medwardson, line 47: perhaps you can use
`post.meal.getMacros()` instead?

medwardson (approved):

app/src/main/java/com/github/se/polyfit/ui/components/postinfo/Text.kt
antomax33: Didn't knew that.

Neb1010 (commented):
Good implementation to have only the most important Meal's infos, and to no have a post 2miles long

PR #218, opened 2024-05-03
Unstable apk

RandomUsername1315:
## PR Description

This PR addresses instability issues in the APK due to two primary factors:

### Minification Setting
The use of `isMinifyEnabled = true` in our Kotlin configuration is intended to enhance app security by reducing the possibility of control-flow attacks. However, this setting seems as a contributing factor to issues during APK generation.

### ViewModel Implementation
The `CreateViewModel` was previously executing database calls directly on the main thread, leading to application freezes and crashes. This process has been modified to use the IO dispatcher, thereby ensuring operations do not block the main thread and improve app stability.

### Small bug fix
Issue with the icon on the bottom bar not being selected when on the  post info screen.

### DIFF
--- a/.github/workflows/apkPipeline.yml
+++ b/.github/workflows/apkPipeline.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - 'release/**'
       - 'main'
+      - 'bugfix/**'
 
 jobs:
   build:
--- a/.github/workflows/apkPipeline.yml
+++ b/.github/workflows/apkPipeline.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - 'release/**'
       - 'main'
-      - 'bugfix/**'
 
 jobs:
   build:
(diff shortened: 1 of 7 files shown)


.github/workflows/apkPipeline.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - 'release/**'
       - 'main'
+      - 'bugfix/**'
```
RandomUsername1315, line 8: This will be deleted before merging. This is just to make sure that the apk is generated correctly 
```suggestion
```

medwardson (approved):
Tested on my emulator and works without any issues

RandomUsername1315:
> LGTM, the app seems stable, I played a bit with it and did not manage to "crash" it.
> 
> But I found probably a weird behavior with the graph if I added a meal for like yesterday or 10 years ago the kcal was added in the graph for today.

Nice find, maybe something that we can add to the sprintbacklog for next time as it does cause the app to crash.

PR #219, opened 2024-05-03
Still draft, I'm not going crazy (maybe)

antomax33:


app/src/main/java/com/github/se/polyfit/ui/screen/OverviewScreen.kt
```diff
@@ -175,6 +175,7 @@ private fun callCamera(
     startCamera: ManagedActivityResultLauncher<Intent, ActivityResult>,
     requestPermissionLauncher: ManagedActivityResultLauncher<String, Boolean>
 ): () -> Unit = {
+  Log.d("OverviewScreen", "Photo button clicked")
```
Neb1010, line 178: I think the Log.d shouldn't be there 

```suggestion
  
```

app/src/androidTest/java/com/github/se/polyfit/end2end/EndTest.kt
```diff
@@ -0,0 +1,164 @@
...
+      GrantPermissionRule.grant(android.Manifest.permission.CAMERA)
+
+  private val overviewViewModel: OverviewViewModel = mockk(relaxed = true)
+
+  @Before
+  fun SettingUp() {
+    mockkStatic(Log::class)
+    System.setProperty("isTestEnvironment", "true")
+
+    every { overviewViewModel.storeMeal(any()) } returns 1L // or any other appropriate return value
```
Neb1010, line 71:
```suggestion
     
```

Mister ChatGPT, copilote or even gemini ?

Neb1010 (approved):
So if recap :
- directly on the overview screen
- take a "picture" for the api
- check the ingredient

Now, the app has more features and overview screen, it would be nice in the future to kind of perform a random walk on the app to check more features.

I saw you had lot of problems doing it so good job you managed to make a end2end test

app/src/androidTest/java/com/github/se/polyfit/end2end/EndTest.kt
medwardson: lol

PR #238, opened 2024-05-06
Recipe dataclass

RandomUsername1315:
# Pull Request Overview

This PR adds the basic backend (both data classes and the classes that handles the storing of `Recipe` on Firebase) for recipe recommendations given a list of ingredients.

## Changes

- **Create `Recipe.kt` Data Class**: This class represents the embodiments of a recipe.

- **Create `RecipeInformation.kt` Data Class**: The reason why this is a separate class from `Recipe` is to follow the same structure as the API (Two separate calls are needed to obtain all the necessary information)

- **Create `RecipeFirebaseRepository.kt` Class**: This class handles all logic related to storing and fetching recipes for a specific user.

## Implementation Checklist

- [x] Recipe
- [x] RecipeInformation
- [x] RecipeFirebaseRepository

## Questions for Reviewers

No Questions

## Future PR

- Modify the Spoonacular API to fetch recipes from a list of potential ingredients.
- Add the UI.
- ViewModel for the UI.

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/model/recipe/Recipe.kt
+++ b/app/src/main/java/com/github/se/polyfit/model/recipe/Recipe.kt
@@ -13,48 +13,39 @@ data class Recipe(
     val recipeInformation: RecipeInformation
 ) {
 
-    fun serialize(): Map<String, Any> {
-        return serialize(this)
-    }
+  fun serialize(): Map<String, Any> {
+    return serialize(this)
+  }
 
-    companion object {
-        fun serialize(recipe: Recipe): Map<String, Any> {
-            return mutableMapOf<String, Any>().apply {
-                this["id"] = recipe.id
-                this["title"] = recipe.title
-                this["imageURL"] = recipe.imageUrl
-                this["usedIngredients"] = recipe.usedIngredients
-                this["missingIngredients"] = recipe.missingIngredients
-                this["likes"] = recipe.likes
-                this["recipeInformation"] = recipe.recipeInformation.serialize()
-            }
-        }
-
+  companion object {
+    fun serialize(recipe: Recipe): Map<String, Any> {
+      return mutableMapOf<String, Any>().apply {
+        this["id"] = recipe.id
+        this["title"] = recipe.title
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/github/se/polyfit/model/recipe/RecipeInformation.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.github.se.polyfit.model.ingredient.Ingredient
+import com.github.se.polyfit.model.nutritionalInformation.MeasurementUnit
+
+/** This is a separate class from the Recipe.kt class to reflect the structure of the api */
+data class RecipeInformation(
+    val vegetarian: Boolean,
+    val vegan: Boolean,
+    val glutenFree: Boolean,
+    val dairyFree: Boolean,
+    val listIngredients: List<Ingredient>,
```
aprilslis, line 13:
```suggestion
    val ingredients: List<Ingredient>,
```

app/src/main/java/com/github/se/polyfit/model/recipe/Recipe.kt
```diff
@@ -0,0 +1,63 @@
...
+
+import android.util.Log
+import java.net.URL
+
+data class Recipe(
+    val id: Long,
+    val title: String,
+    val imageUrl: URL,
+    val usedIngredients: Long,
+    val missingIngredients: Long,
```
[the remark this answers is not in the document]
RandomUsername1315: This is done to represent the fields of the API as much as closely as possible ([Spoonacular API](https://rapidapi.com/spoonacular/api/recipe-food-nutrition)). Essentially, it includes the ingredients that the user has (`usedIngredients`) and the ingredients that are missing (`missingIngredients`) for a specific recipe. I thought this could be useful for future features.

app/src/main/java/com/github/se/polyfit/model/recipe/RecipeInformation.kt
RandomUsername1315: Yeah you're right that a way better naming scheme.

PR #239, opened 2024-05-07
Main Setting Screen

aprilslis:
This is the UI for the main settings screen. From here, people will be able to navigate to specific setting pages. Those pages will be made in following PRs so everything is small.

Screenshot 2024-05-07 at 7 25 12 PM

### DIFF
--- a/app/src/androidTest/java/com/github/se/polyfit/ui/screen/SettingsTest.kt
+++ b/app/src/androidTest/java/com/github/se/polyfit/ui/screen/SettingsTest.kt
@@ -41,7 +41,7 @@ class SettingsTest : TestCase() {
 
   @Test
   fun everythingIsDisplayed() {
-    ComposeScreen.onComposeScreen<SettingsScreen>(composeTestRule) {
+    ComposeScreen.onComposeScreen<SettingScreen>(composeTestRule) {
       title {
         assertExists()
         assertIsDisplayed()
--- a/app/src/androidTest/java/com/github/se/polyfit/ui/screen/SettingsTest.kt
+++ b/app/src/androidTest/java/com/github/se/polyfit/ui/screen/SettingsTest.kt
@@ -1,0 +1,70 @@
+package com.github.se.polyfit.ui.screen
+
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.automirrored.outlined.ExitToApp
+import androidx.compose.material.icons.outlined.AccountCircle
+import androidx.compose.material.icons.outlined.Build
+import androidx.compose.material.icons.outlined.Info
+import androidx.compose.material.icons.outlined.Lock
+import androidx.compose.ui.test.assertCountEquals
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onAllNodesWithTag
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.github.se.polyfit.model.settings.SettingOption
+import io.github.kakaocup.compose.node.element.ComposeScreen
+import io.mockk.mockk
+import io.mockk.verify
+import junit.framework.TestCase
+import kotlin.test.Test
+import org.junit.Before
+import org.junit.Rule
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class SettingsTest : TestCase() {
+  @get:Rule val composeTestRule = createComposeRule()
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/polyfit/ui/screen/SettingsScreen.kt
```diff
@@ -0,0 +1,93 @@
...
+  Column(
+      modifier = modifier.fillMaxSize().testTag("SettingsScreen"),
+      horizontalAlignment = Alignment.CenterHorizontally,
+      verticalArrangement = Arrangement.Top) {
+        Text(
+            text = context.getString(R.string.settings),
+            style = MaterialTheme.typography.headlineMedium,
+            modifier = Modifier.padding(48.dp).testTag("SettingsTitle"))
+
+        LazyColumn(modifier = Modifier.testTag("SettingsList")) {
```
medwardson, line 39: im not sure it's strictly neccesary (since we won't ever have an empty setting screen), but maybe you can put a default text if there's no settings passed

app/src/main/res/values/strings.xml
```diff
@@ -32,4 +32,9 @@
     <string name="newPostTitle">"New Post"</string>
     <string name="newPostDescriptionPlaceholder">" What's on your mind today?"</string>
     <string name="postButton">"Post"</string>
+    <string name="account">"Account"</string>
+    <string name="termsAndConditions">"Terms &amp; Conditions"</string>
+    <string name="privacyPolicy">"Privacy Policy"</string>
+    <string name="about">"About"</string>
+    <string name="logout">"Logout"</string>
```
medwardson, line 39: i see these aren't used yet, is it for your next PR

app/src/androidTest/java/com/github/se/polyfit/ui/screen/SettingsTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+      divider {
+        assertExists()
+        assertIsDisplayed()
+      }
+
+      composeTestRule.onAllNodesWithTag("SettingItem").assertCountEquals(5)
+    }
+  }
+}
```
medwardson, line 70: if you do add a default text, would be good to have a test

medwardson (approved):
Nice job and thanks for keeping it short, makes it really easy to review. Looks good, left minor question

app/src/main/java/com/github/se/polyfit/ui/screen/SettingsScreen.kt
Neb1010: I agree with Mesha, not sure about the use of a lazyColumn, just a Column should be good

Neb1010 (approved):
Good job for you work!, this is clean and easy to review !

sharif-bbd (approved):
Really good job for the UI !

app/src/main/java/com/github/se/polyfit/ui/screen/SettingsScreen.kt
```diff
@@ -0,0 +1,93 @@
...
+// @Preview
+// @Composable
+// fun PreviewSettingsScreen() {
+//  val settings: List<SettingOption> = listOf(
+//    SettingOption("Account", Icons.Outlined.AccountCircle, {}),
+//    SettingOption("Terms & Conditions", Icons.Outlined.Build, {}),
+//    SettingOption("Privacy Policy", Icons.Outlined.Lock, {}),
+//    SettingOption("About", Icons.Outlined.Info, {}),
+//    SettingOption("Logout", Icons.AutoMirrored.Outlined.ExitToApp, {}),
+//  )
```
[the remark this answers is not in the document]
aprilslis: I don't think a viewmodel makes sense here, as it is not dynamic. But, I will create a default list, not in this PR, this is going to be done in the next PR

app/src/main/res/values/strings.xml
aprilslis: Yes, it will be used when I create the list of setting optiosn

app/src/main/java/com/github/se/polyfit/ui/screen/SettingsScreen.kt
aprilslis: Thats fair! I think it should be the same either way, and if we get too many settings it would be good to have a lazy column? I will keep it as is for now, but if we have functionality issue we can address later

antomax33 (approved):
LGTM

PR #245, opened 2024-05-08
Feature/UI/recipe recommendation

RandomUsername1315:
# Pull Request Overview

This is the first PR in a series of two for the interface for recipe recommendation. It introduces a list of recipes and gives the option to add a recipe to the user's saved recipes.

## Changes

- **Create `LikeButton.kt`**: This shows the like count sourced from the Spoonacular API.
- **Create `RecipeRecommendationScreen.kt` Data Class**: The screen integrates all the components created by this PR.
- **Create `BookmarkButton.kt` Class**: This button lets users save their favorite recipes.
- **Create `RecipeCard.kt` Class**: Represents a single graphic of recipe instance.

## Notes for Reviewers

As this is my first UI task for this project, I'd really appreciate any detailed and honest feedback you can give.

## Future Work

- Add a detailed view when a recipe is selected.
- Implement the view model to handle the backend of the UI.


app/src/main/java/com/github/se/polyfit/ui/components/button/LikeButton.kt
```diff
@@ -0,0 +1,56 @@
...
+fun LikeButton(
+    likesCount: Long,
+    modifier: Modifier = Modifier,
+    onLikeClicked: (Boolean) -> Unit = {}
+) {
+  FilledTonalButton(
+      onClick = { onLikeClicked },
+      colors = ButtonDefaults.buttonColors(containerColor = Color.White),
+      shape = RoundedCornerShape(20.dp),
+      enabled = false, // Set this to true
```
antomax33, line 33: Say why you disable this button instead of chatgpt comment

app/src/main/java/com/github/se/polyfit/ui/components/button/LikeButton.kt
```diff
@@ -0,0 +1,56 @@
...
+      }
+}
+
+@Preview
+@Composable
+fun PreviewCustomLikeButton() {
+  LikeButton(
+      likesCount = 6113,
+  )
+}
```
antomax33, line 56:
```suggestion
/*@Preview
@Composable
fun PreviewCustomLikeButton() {
  LikeButton(
      likesCount = 6113,
  )
}*/
```

app/src/main/java/com/github/se/polyfit/ui/components/recipe/RecipeCard.kt
```diff
@@ -0,0 +1,82 @@
...
+              modifier =
+                  Modifier.align(Alignment.BottomStart).padding(horizontal = 8.dp, vertical = 4.dp))
+
+          LikeButton(recipe.likes, modifier = Modifier.align(Alignment.TopEnd))
+
+          BookmarkButton(
+              recipe, onBookmarkClick, onBookmarkRemove, Modifier.align(Alignment.BottomEnd))
+        }
+      }
+}
```
antomax33, line 76: Too much nesting, you can create a function foregroundRecipe with everything inside the box.

app/src/main/java/com/github/se/polyfit/ui/components/recipe/RecipeCard.kt
```diff
@@ -0,0 +1,82 @@
...
+
+@Preview
+@Composable
+fun preview() {
+  RecipeCard(recipe = Recipe.default())
+}
```
antomax33, line 82:
```suggestion

/*@Preview
@Composable
fun preview() {
  RecipeCard(recipe = Recipe.default())
}*/
```

app/src/main/java/com/github/se/polyfit/ui/screen/RecipeRecommendationScreen.kt
```diff
@@ -0,0 +1,46 @@
...
+        }
+        recipes.forEach { item { RecipeCard(recipe = it) } }
+      }
+}
+
+@Composable
+@Preview
+fun recipeDisplayPreview() {
+  recipeDisplay(recipes = listOf(Recipe.default(), Recipe.default()))
+}
```
antomax33, line 46:
```suggestion
/*@Composable
@Preview
fun recipeDisplayPreview() {
  recipeDisplay(recipes = listOf(Recipe.default(), Recipe.default()))
}*/
```

app/src/main/java/com/github/se/polyfit/ui/screen/RecipeRecommendationScreen.kt
```diff
@@ -0,0 +1,46 @@
...
+
+@Composable
+fun recipeDisplay(recipes: List<Recipe>) {
+
+  LazyColumn(
+      horizontalAlignment = Alignment.CenterHorizontally,
+      modifier = Modifier.testTag("RecipeList")) {
+        item {
+          Text(
+              text = "Recommended Recipes",
```
antomax33, line 33: It would be great to have the text in the string.xml file. You can take Local.context to get the context

app/src/main/java/com/github/se/polyfit/ui/components/button/BookmarkButton.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.github.se.polyfit.model.recipe.Recipe
+
+@Composable
+fun BookmarkButton(
+    recipe: Recipe,
+    addRecipe: (Recipe) -> Unit = {},
+    removeRecipe: (Recipe) -> Unit = {},
+    modifier: Modifier = Modifier,
+) {
+  var isClicked by remember { mutableStateOf(false) }
```
medwardson, line 26: maybe for consistency with the `IconToggleButton`, a more suitable name would be `isChecked`

app/src/main/java/com/github/se/polyfit/ui/components/button/BookmarkButton.kt
```diff
@@ -0,0 +1,45 @@
...
+      onCheckedChange = {
+        isClicked = !isClicked
+        if (isClicked) {
+          addRecipe(recipe)
+        } else {
+          removeRecipe(recipe)
+        }
+      },
+      modifier = modifier.testTag("BookmarkButton"),
+      colors = IconButtonDefaults.filledIconToggleButtonColors(containerColor = Color(0xFFE1BEE7)),
```
medwardson, line 38: Maybe add this as a variable in `Color.kt` so it can be reused in the future if anything have the same color (and so that it provides a good description)

app/src/main/java/com/github/se/polyfit/ui/components/button/LikeButton.kt
```diff
@@ -0,0 +1,56 @@
...
+    likesCount: Long,
+    modifier: Modifier = Modifier,
+    onLikeClicked: (Boolean) -> Unit = {}
+) {
+  FilledTonalButton(
+      onClick = { onLikeClicked },
+      colors = ButtonDefaults.buttonColors(containerColor = Color.White),
+      shape = RoundedCornerShape(20.dp),
+      enabled = false, // Set this to true
+      modifier = modifier.padding(horizontal = 25.dp, vertical = 8.dp).testTag("LikeButton")) {
```
medwardson, line 34: It's usually suggested in design/ui to follow an 8 point grid (https://uxplanet.org/everything-you-should-know-about-8-point-grid-system-in-ux-design-b69cb945b18d), so its best to use multiples of 8.
```suggestion
      modifier = modifier.padding(horizontal = 24.dp, vertical = 8.dp).testTag("LikeButton")) {
```

app/src/main/java/com/github/se/polyfit/ui/components/button/LikeButton.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+@Composable
+fun LikeButton(
+    likesCount: Long,
```
medwardson, line 25:
```suggestion
    likeCount: Long,
```

medwardson (commented):
Charles doing UI

app/src/main/java/com/github/se/polyfit/ui/screen/RecipeRecommendationScreen.kt
RandomUsername1315: I always forget to remove the previews. Thanks for reminding me to.

app/src/main/java/com/github/se/polyfit/ui/components/recipe/RecipeCard.kt
RandomUsername1315: Yep, you're completely right; it's almost unreadable. Thanks for catching that. Do you think it's okay if I remove the box, or do you think I should modularize even further down @antomax33 ?

app/src/main/java/com/github/se/polyfit/ui/screen/RecipeRecommendationScreen.kt
RandomUsername1315: Is the strings.xml file the one that contains all the text displayed on the app? And should all the text we want to display be in there instead of hardcoded?

app/src/main/java/com/github/se/polyfit/ui/components/button/LikeButton.kt
RandomUsername1315: Oh, I never knew that. Thanks for linking an article; I'll read it right now!

app/src/main/java/com/github/se/polyfit/ui/components/button/BookmarkButton.kt
RandomUsername1315: Nice, I didn't know that this existed. Thanks for pointing it out, I'll use it in the future.

app/src/main/java/com/github/se/polyfit/ui/components/button/BookmarkButton.kt
RandomUsername1315: Yep, way better naming. Thanks for the suggestion.

app/src/main/java/com/github/se/polyfit/ui/components/recipe/RecipeCard.kt
Neb1010: modularization of the AsyncImage and Text, something like this should be enough, the LikeButton is 1 ligne so...

```
@Composable
fun RecipeImage(recipe: Recipe, onCardClick: (Recipe) -> Unit){
....
}

@Composable
fun RecipeText(text: String){
....
}
```

app/src/main/java/com/github/se/polyfit/ui/screen/RecipeRecommendationScreen.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.compose.ui.text.font.FontWeight
+import androidx.core.content.ContextCompat
+import androidx.navigation.NavHostController
+import com.github.se.polyfit.R
+import com.github.se.polyfit.model.recipe.Recipe
+import com.github.se.polyfit.ui.components.GenericScreen
+import com.github.se.polyfit.ui.components.recipe.RecipeCard
+
+@Composable
+fun RecipeRecommendationScreen(navController: NavHostController, recipes: List<Recipe>) {
```
Neb1010, line 20: You shouldn't pass directly the navController, but a function of it. [More explication](https://developer.android.com/guide/navigation/use-graph/navigate)

Neb1010 (approved):
Good job but you went to the dark side of the force

app/src/main/java/com/github/se/polyfit/ui/components/recipe/RecipeCard.kt
RandomUsername1315: Okay, thanks. I was able to refactor the images, but the text needs to directly be in a box because it uses the function `.align`.

app/src/main/java/com/github/se/polyfit/ui/screen/RecipeRecommendationScreen.kt
RandomUsername1315: The generic screens need a `navController`. Yeah, directly passing it is not ideal. We should fix this in a later PR, if we have time. I'll need to pass navigation functions in the implementation that creates the view model, and I will follow your advice.

PR #246, opened 2024-05-08
add recipe button

antomax33:


app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
```diff
@@ -46,7 +47,8 @@ fun NutritionScreen(
         BottomBar(
             setMeal = mealViewModel::setMeal,
             isComplete = isComplete,
-            navigateForward = navigateForward)
+            navigateForward = navigateForward,
+            mealViewModel = mealViewModel)
```
aprilslis, line 51: I think it is generally better to pass the method only (see `setMeal` above) instead of the whole mealViewModel
```suggestion
            updateMealData = mealViewModel::updateMealData)
```

app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
```diff
@@ -79,14 +81,26 @@ private fun TopBar(navigateBack: () -> Unit) {
 }
 
 @Composable
-private fun BottomBar(setMeal: () -> Unit, isComplete: Boolean, navigateForward: () -> Unit) {
+private fun BottomBar(
+    setMeal: () -> Unit,
+    isComplete: Boolean,
+    navigateForward: () -> Unit,
+    mealViewModel: MealViewModel
```
aprilslis, line 88:
```suggestion
    updateMealData: () -> Unit
```

aprilslis (commented):
Looks good but one small thing

app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
aprilslis: can you do something like `updateMealData = { createdAt ->  mealViewmodel.updateMealData(createdAt = createdAt) }`

app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
```diff
@@ -79,14 +81,26 @@ private fun TopBar(navigateBack: () -> Unit) {
...
             horizontalAlignment = Alignment.CenterHorizontally) {
               PrimaryButton(
-                  onClick = { Log.v("Add Recipe", "Clicked") },
+                  onClick = {
+                    Log.v("Add Recipe", "Clicked")
+                    navigateForward()
+                    // Set the date of the meal to the minimum value to set a default value
+                    // who will not be on the data we see.
+                    updateMealData(LocalDate.MIN)
+                    setMeal()
```
RandomUsername1315, line 102: ## Issue Description

There appears to be a problem when the meal is incomplete. The button `Add To Diary` remains always active even when not all the necessary information has been entered.. This issue allows the `setMeal()` function to be called prematurely, which leads to an error. The function is designed to check if the meal information is complete before it proceeds to store the data in the database.

### Example Of the Error
[Watch the example video](https://github.com/swent-group10/polyfit/assets/100274930/9966ae59-0916-4912-a4ff-19c67d88999d)

RandomUsername1315 (commented):
Everything looks ok. Just one small issue.

medwardson (approved):
Looks like you've addressed all the comments, works as expected, nice job

PR #253, opened 2024-05-09
Modified Spoon API to fetch recommendations

RandomUsername1315:
# Pull Request Overview

- This PR builds upon #245 by adding backend logic to recommend recipes based on a potential list of ingredients. It retrieves information about the recipes, such as the number of missing ingredients, the number of likes, and an image of the recipe.

## Changes

- **Create `RecipeFromIngredientsResponseAPI.kt` class**: Represents the potential response from the API for a recipe-from-ingredients query.
- **Modify `SpoonacularApiCaller.kt` class**: Updated to handle recommendation requests.
- **Modify `RecipeCard.kt` class**: Updated the placeholder image to the logo.
- **Create `RecipeRecommendationViewModel.kt` class**: Represents the ViewModel for the recipe recommendation UI.

## Future Work

- Add a detailed view when a recipe is selected.

## Video of the New Implementation
[View the implementation video](https://github.com/swent-group10/polyfit/assets/100274930/7e1248ed-7966-4eb2-9de8-c51c392cce64)


app/src/main/java/com/github/se/polyfit/data/api/RecipeFromIngredientsResponseAPI.kt
```diff
@@ -0,0 +1,48 @@
...
+) {
+  companion object {
+    fun faillure(): RecipeFromIngredientsResponseAPI {
+      return RecipeFromIngredientsResponseAPI(status = APIResponse.FAILURE, recipes = emptyList())
+    }
+
+    fun fromJsonObject(jsonArray: JSONArray): RecipeFromIngredientsResponseAPI {
+      try {
+        val recipes = mutableListOf<Recipe>()
+        for (i in 0 until jsonArray.length()) {
```
Neb1010, line 21: 1 pt of complexity

app/src/main/java/com/github/se/polyfit/viewmodel/recipe/RecipeRecommendationViewModel.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.coroutines.withContext
+
+@HiltViewModel
+class RecipeRecommendationViewModel
+@Inject
+constructor(private val spoonacularApiCaller: SpoonacularApiCaller) : ViewModel() {
+
+  suspend fun recipeFromIngredients(ingredients: List<String>): List<Recipe> {
+    val recipesResponse =
+        withContext(Dispatchers.IO) { spoonacularApiCaller.recipeByIngredients(ingredients) }
```
Neb1010, line 18: Shouldn't be the default ? I remember we talked about it, and you told me to let `.DEFAULT` and it will pick the best thread.

Neb1010 (approved):
Just the part of the `NavController` that's not perfect about the [doc](https://developer.android.com/guide/navigation/use-graph/navigate), but it's maybe an old pr, where the person didn't know about it, and to refactor this would take a bit of time. In the doc they don't precise what happend if we do it. If the future APK is unstable and there's some bug, maybe a good idea too look at it

app/src/main/java/com/github/se/polyfit/MainActivity.kt
```diff
@@ -75,6 +76,10 @@ class MainActivity : ComponentActivity() {
                 navigateBack = navigation::goBack, navigateTo = navigation::navigateToAddMeal)
           }
           composable(Route.AddMeal) { AddMealFlow(navigation::goBack, navigation::navigateToHome) }
+
+          composable(Route.RecipeRecommendation) {
+            RecipeRecommendationScreen(navController = navController)
```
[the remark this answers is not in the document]
RandomUsername1315: I completely agree with you. However, since I am using the pre-implemented component `GenericScreen.kt`, which directly uses the `navController`, I can't do much without refactoring this component. This should be done in another PR. If you have some free time, feel free to refactor it.

app/src/main/java/com/github/se/polyfit/ui/screen/RecipeRecommendationScreen.kt
```diff
@@ -4,30 +4,43 @@ import androidx.compose.foundation.lazy.LazyColumn
...
 import com.github.se.polyfit.R
 import com.github.se.polyfit.model.recipe.Recipe
 import com.github.se.polyfit.ui.components.GenericScreen
 import com.github.se.polyfit.ui.components.recipe.RecipeCard
+import com.github.se.polyfit.viewmodel.recipe.RecipeRecommendationViewModel
 
 @Composable
-fun RecipeRecommendationScreen(navController: NavHostController, recipes: List<Recipe>) {
+fun RecipeRecommendationScreen(
+    navController: NavHostController,
```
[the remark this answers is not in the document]
RandomUsername1315: Same as before, because it uses the component `GenericScreen.kt` there isn't much I can do.

app/src/main/java/com/github/se/polyfit/data/api/RecipeFromIngredientsResponseAPI.kt
RandomUsername1315: Out of 50 so I should be ok

app/src/main/java/com/github/se/polyfit/viewmodel/recipe/RecipeRecommendationViewModel.kt
RandomUsername1315: You're completely right. Thank you for catching that !

PR #255, opened 2024-05-10
 buttons component and nutritionalInformation

sharif-bbd:
This branch aims to refactor the components package to lower the complexity. 

# files to modify:

- [x] BookmarkButton.kt
- [x] GradientButton.kt
- [x] NutritionalInformation.kt
- [x] MealTagDialog.kt
 Since the requirements on complexity changed and it isn't necessary anymore I didn't continue the refactoring of the components package but can and will in the future whenever I get the time to do so


app/src/main/java/com/github/se/polyfit/ui/components/button/BookmarkButton.kt
```diff
@@ -29,18 +29,26 @@ fun BookmarkButton(
...
   }
 }
+
+private fun addOrRemoveRecipe(
+    isChecked: Boolean,
+    recipe: Recipe,
+    addRecipe: (Recipe) -> Unit = {},
+    removeRecipe: (Recipe) -> Unit = {}
+) {
+  when (isChecked) {
```
antomax33, line 50: Now that we don't need <50 complexity. Isn't it better to just put an if else?

antomax33 (approved):
It is not more readable but it was good for the complexity

app/src/main/java/com/github/se/polyfit/ui/components/button/BookmarkButton.kt
sharif-bbd: Well i'm thinking it is good practice and if for the next milestone they are stricter on complexity it will be changed anyway

PR #257, opened 2024-05-10
Feature/login/keepconnected

antomax33:


app/src/main/java/com/github/se/polyfit/ui/utils/Authentication.kt
```diff
@@ -4,59 +4,85 @@ import android.app.Activity
...
+
+    val signInIntent =
+        AuthUI.getInstance().createSignInIntentBuilder().setAvailableProviders(providers).build()
+
+    signInLauncher!!.launch(signInIntent)
+  }
+
+  fun setCallback(callback: () -> Unit, justToHave2ArguementHilt: Int) {
+    this.callback = callback
+  }
```
antomax33, line 57: I need to add a useless argument otherwise Hilt will be confuse with another function

app/src/androidTest/java/com/github/se/polyfit/MainActivityTest.kt
Neb1010, line 1: ### Strengths
- Usage of ActivityScenario: ActivityScenario is well-used to launch and interact with the MainActivity.
- Clear Test Logic: Each test has a clear purpose: testing sign-in and sign-out functionalities.
- Logging: Logging is added to trace the waiting period for authentication, which can be useful for debugging.
### Improvement
- Busy Waiting Loop: The while loop with sleep(10) is a busy-waiting loop, which is not efficient and can slow down the tests.
- Timeout Handling: The current timeout handling is a hardcoded loop with a limit of 10,000 iterations, which might not be reliable.
- Callback Usage: The setCallback method is called with empty logic {}, which might not be effective for testing.

app/src/androidTest/java/com/github/se/polyfit/ui/utils/AuthenticationTest.kt
Neb1010, line 1: ### strengths 
- Global Timeout Rule: The inclusion of a global timeout rule ensures that tests do not run indefinitely, which is good practice.
- Setup and Teardown Methods: The @BeforeTest and @AfterTest methods are used to prepare the test environment and clean up afterwards.
### Improvement
- Commented-Out Code: Several lines of code are commented out, which makes it unclear what the intended behavior is.

app/src/main/java/com/github/se/polyfit/MainActivity.kt
```diff
@@ -40,21 +48,31 @@ class MainActivity : ComponentActivity() {
...
+            userFirebaseRepository = userFirebaseRepository)
+
     // TODO: technical debt, next deadline find better way to pass arguments from overview screen
     // to add meal screen
     setContent {
       PolyfitTheme {
         val navController = rememberNavController()
         val navigation = Navigation(navController)
-        NavHost(navController = navController, startDestination = Route.Register) {
+        authentication.setCallback({ navigation.navigateToHome() }, 3)
```
Neb1010, line 64:
```
        authentication.setCallback({ navigation.navigateToHome() }, 3)
```
magic number

app/src/main/java/com/github/se/polyfit/ui/utils/Authentication.kt
Neb1010: This function has to be named like this ?

app/src/main/java/com/github/se/polyfit/ui/utils/Authentication.kt
Neb1010, line 1: ### Strengths
- Comprehensive Functionality: The class covers all essential authentication tasks, including sign-in, sign-out, and user information management.
- Separation of Concerns: Methods are well-separated, each handling specific aspects of the authentication process.
### Improvement
- Callback Initialization: The callback initialization message ("Callback not set") could be misleading. Consider a more informative default or a way to ensure it is always set.
- Null Checks: Additional null checks and safety mechanisms could be added to avoid potential null pointer exceptions.
- Context Management: The use of context should be consistent and preferably final to avoid accidental reassignment.
- ui/utils is maybe no the right place for this files

Neb1010 (approved):
Very good addition to the project.

medwardson:
I'm not sure if this is expected, but, it seems like a bit of a weird UX. When I open the app for the first time, it doesn't let me pretty the sign-in button. Rather, it automatically presses it and gives the login message. The expected experience would be that it only prompts once I click the button

antomax33:
@medwardson If it is the first time the application is launched, the first page will be shown and the user need to click on the login button, otherwise, the application will auto-loged (probably your case) and directly get the information and go to the homepage.

medwardson:
I am talking about the first time it is launched after both clear data and uninstall, it auto presses the button

> If it is the first time the application is launched, the first page will be shown and the user need to click on the login button

This part does not happen for me, I immediately go here:

Screenshot 2024-05-20 at 10 48 11 AM

antomax33:
@medwardson Yeah even if you uninstall and clear cache and everything, google connect will keep in memory that you have been loged to this app with your google account. So it will auto-login. There is no difference betweeen this and if you go to the app normally.

app/src/main/java/com/github/se/polyfit/di/UserModule.kt
```diff
@@ -90,16 +89,6 @@ object UserModule {
...
-  @Provides
-  @Singleton
-  fun provideAuthentication(
-      @ApplicationContext context: Context,
-      user: User,
-      userFirebaseRepository: UserFirebaseRepository
-  ): AuthenticationCloud {
-    return AuthenticationCloud(context, user, userFirebaseRepository)
-  }
-
```
RandomUsername1315, line 102: Potential issue I just realized: the first time the user signs into the app, because the value of `user` is set to `"test"` when Firebase is initialized, the meals will be stored there instead of in the current user's folder. Do you think this could potentially cause problems ?

app/src/main/java/com/github/se/polyfit/model/user/User.kt
```diff
@@ -26,7 +26,7 @@ data class User(
   }
 
   fun isSignedIn(): Boolean {
-    return id.isNotEmpty()
+    return id.isNotEmpty() && !id.equals("testUserID")
```
RandomUsername1315, line 29: Would it be more consistent to also check if the Firebase user is logged in?

RandomUsername1315 (commented):
Really good overall just a few remarks.

app/src/main/java/com/github/se/polyfit/ui/utils/Authentication.kt
```diff
@@ -4,59 +4,84 @@ import android.app.Activity
...
 
-    signInLauncher.launch(signInIntent)
+    signInLauncher!!.launch(signInIntent)
+  }
+
+  fun signOut() {
+    AuthUI.getInstance().signOut(context)
+    user.signOut()
+    isAnswered = false
   }
```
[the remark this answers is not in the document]
aprilslis: My code has been merged, so @antomax33 if theres anything that is not addressed in the signout in my implementation you can include it to work with keepconnected

PR #263, opened 2024-05-14
Remove the dark mode because we don't have one for the moment.

antomax33:
As the coach said, we can remove the dark mode. Way easier than updating all the application to work on dark-mode

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/ui/theme/Theme.kt
+++ b/app/src/main/java/com/github/se/polyfit/ui/theme/Theme.kt
@@ -1,17 +1,13 @@
 package com.github.se.polyfit.ui.theme
 
 import android.app.Activity
-import android.os.Build
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.material3.MaterialTheme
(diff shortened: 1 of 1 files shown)


medwardson (commented):
It looks good, my only question is why not delete dark theme parameter in PolyfitTheme and set `WindowCompat.getInsetsController(window, view).isAppearanceLightStatusBars = darkTheme` to be `WindowCompat.getInsetsController(window, view).isAppearanceLightStatusBars = false` instead?

app/src/main/java/com/github/se/polyfit/ui/theme/Theme.kt
```diff
@@ -37,18 +30,9 @@ private val LightColorScheme =
 fun PolyfitTheme(
     darkTheme: Boolean = isSystemInDarkTheme(),
```
medwardson, line 31:
```suggestion
```

medwardson (commented):
It seems like the variable is still there

medwardson (approved):
Nice, thanks for the fix

PR #264, opened 2024-05-14
Add images to posts

RandomUsername1315:


app/src/main/java/com/github/se/polyfit/ui/components/selector/PictureSelector.kt
```diff
@@ -1,30 +1,97 @@
...
+
+  val requestPermission =
+      rememberLauncherForActivityResult(ActivityResultContracts.RequestPermission()) {
+          isGranted: Boolean ->
+        if (isGranted) {
+          takePicturePreview.launch(null) // No input needed, directly launches camera for a preview
+        } else {
+          Toast.makeText(context, PERMISSION_DENIED_MESSAGE, Toast.LENGTH_SHORT).show()
+        }
+      }
```
RandomUsername1315, line 65: Question for reviewers : do you think that this should stay here or be moved to the viewmodel?

app/src/main/java/com/github/se/polyfit/ui/components/selector/PictureSelector.kt
```diff
@@ -1,30 +1,92 @@
...
+import androidx.compose.ui.graphics.asImageBitmap
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.unit.dp
+import androidx.core.content.ContextCompat
 import com.github.se.polyfit.ui.theme.PurpleGrey80
+import com.github.se.polyfit.viewmodel.post.CreatePostViewModel
+
+private const val PERMISSION_DENIED_MESSAGE = "Permission denied"
```
medwardson, line 33: instead of a const here, we can use the string resources

app/src/main/java/com/github/se/polyfit/ui/components/selector/PictureSelector.kt
```diff
@@ -1,30 +1,92 @@
...
 
 @Composable
-fun PictureSelector(modifier: Modifier = Modifier) {
-  // Not sure if we are going to use a picture for the post, so this is just a placeholder
+fun PictureSelector(modifier: Modifier = Modifier, postViewModel: CreatePostViewModel) {
```
medwardson, line 36: Perhaps instead of passing the whole viewmodel, we can just pass `getBitMap` and `setBitMap`

app/src/main/java/com/github/se/polyfit/ui/components/selector/PictureSelector.kt
```diff
@@ -1,30 +1,92 @@
...
-              Text("Picture Placeholder")
+              if (bitmapPicture.value != null) {
+                // Display the bitmap here
+                Image(
+                    bitmap = bitmapPicture.value!!.asImageBitmap(),
+                    contentDescription = "Captured image",
+                    modifier = Modifier.fillMaxSize(),
+                    contentScale = ContentScale.Crop)
+              } else {
+                Text("Tap to take photo")
```
medwardson, line 88: These two can also be moved to the string resources

medwardson (commented):
Looks good! I left some small comments. My only big concern is that from what I could tell testing on my emulator, there is no way to clear the picture. You can take a new picture, but can not remove it altogether. This may be considered a bad user experience, thoughts?

> Question for reviewers : do you think that this should stay here or be moved to the ViewModel?
While I think it's a big chunk of code, it's in line with what we are doing with login, so, I think it's fine to leave here for now.

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
```diff
@@ -98,6 +110,14 @@ data class Post(
       }
     }
 
+    private fun serializeLocalDate(data: LocalDate): Map<String, Any> {
```
medwardson, line 113: I actually think this and desrializeLocalDate could be extracted into a separate class as it can be used by other areas of the app

app/src/androidTest/java/com/github/se/polyfit/model/post/PostTest.kt
```diff
@@ -23,7 +23,7 @@ class PostTest {
   }
 
   @Test
-  fun `getCarbs returns correct nutrient when present`() {
+  fun getCarbsReturnsCorrectNutrientWhenPresent() {
```
aprilslis, line 26: Maybe these changes could be done in a separate PR

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
```diff
@@ -22,7 +23,8 @@ data class Post(
     override var meal: Meal,
     override var createdAt: LocalDate,
     var listOfImages: List<Bitmap> = emptyList(),
-    var listOfURLs: List<StorageReference> = emptyList()
+    var listOfURLs: List<StorageReference> = emptyList(),
+    var imageDownloadURL: Uri? = null
```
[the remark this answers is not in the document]
aprilslis: I would say that null is ok, since empty Uri means "" but if theres no Uri present its better to be clear and say null

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
```diff
@@ -75,9 +79,17 @@ data class Post(
         val description = data["description"] as? String ?: ""
         val location = Location.deserialize(data["location"] as Map<String, Any>)
         val meal = Meal.deserialize(data["meal"] as Map<String, Any>)
-        val createdAt = deserializeLocalDate(data, "createdAt") ?: LocalDate.now()
+        val createdAt = deserializeLocalDate(data, "createdAt")
+        var imageDownloadURL: Uri? = null
 
-        val newPost = Post(userId, description, location, meal, createdAt)
+        if (data["imageDownloadURL"] is String &&
```
aprilslis, line 85: If we are storing it as either `""` or the string, wouldn't this always be true, since in both case it is a string

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
```diff
@@ -77,8 +88,17 @@ constructor(
   }
 
   fun setPost() {
-    runBlocking {
+    viewModelScope.launch {
```
aprilslis, line 91: Good idea!

aprilslis (commented):
Great job!

app/src/main/java/com/github/se/polyfit/ui/components/selector/PictureSelector.kt
```diff
@@ -1,30 +1,92 @@
...
+
+private const val PERMISSION_DENIED_MESSAGE = "Permission denied"
 
 @Composable
-fun PictureSelector(modifier: Modifier = Modifier) {
-  // Not sure if we are going to use a picture for the post, so this is just a placeholder
+fun PictureSelector(modifier: Modifier = Modifier, postViewModel: CreatePostViewModel) {
+  val context = LocalContext.current
+
+  var bitmapPicture = remember { mutableStateOf(postViewModel.getBitMap()) }
```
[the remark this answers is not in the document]
RandomUsername1315: Good catch. I will make it `var .... by remember` to make even more explicit that it can be modified.

app/src/main/java/com/github/se/polyfit/ui/components/selector/PictureSelector.kt
RandomUsername1315: Yep did not think of that. That is way cleaner. Thanks for the advice !

app/src/androidTest/java/com/github/se/polyfit/model/post/PostTest.kt
RandomUsername1315: Due to the post class storing a `Uri` and the fact that this class does not exist in the unit tests, the test file had to be moved to the `androidTest` folder. In this folder, you cannot have function names with apostrophes or spaces, so all the function names in that file had to be renamed.

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
RandomUsername1315: LocalDate is a class directly provided by the Kotlin library. Locally, it can be directly used as a map due to its structure. However, when uploading `LocalDate` as is to Firebase, we encountered strange issues. Firebase tried to convert it into its own type, and when deserializing the post, the program crashed due to an unexpected type. In order to have symmetric serialization and deserialization, these classes were created to store it. Those functions should not be used in theory ouside of deserialization and serialization with the intent to store in on the cloud.

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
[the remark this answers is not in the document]
RandomUsername1315: Yeah, but the issue with using `Uri.EMPTY` is that you cannot use it in unit tests due to some weird behavior where it thinks it is a pointer. The test has to be moved to the Android folder, which is what I just did. I think that using it makes the code clearer.

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
RandomUsername1315: No for some reason meal uses a complete different serialization method. Which I cannot remember why. It would be a great idea to utilize the same methods everywhere. But I think that it is outside the scope of this PR. What do you think ?

RandomUsername1315:
@medwardson Totally agree with you about the UI. I tried to follow the Figma as closely as possible. So, the way I see it is that we have two options. The first one is to add a new button with an 'x'. The other is if there is already a picture in the box, then clicking on it removes the picture.

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
RandomUsername1315: Good catch !

app/src/main/java/com/github/se/polyfit/model/post/Post.kt
RandomUsername1315: I mean it is a 2 lines change.

medwardson: Yeah, can be done outside of this PR :D

medwardson:
@RandomUsername1315 I think of those two, the 'x' will be a better UX since if they accidentally click the picture they'll lose the one they took

RandomUsername1315:
@medwardson what do you think of this ?

https://github.com/swent-group10/polyfit/assets/100274930/d6567155-0558-4038-bb47-b65d8b5adf41

aprilslis (approved):
Thanks for addressing the changes, nice job

PR #267, opened 2024-05-14
Edit Meal from Daily Recap

medwardson:
## Feature: Edit Meals
Users can click on any of the meals while viewing their history in the daily recap to update the data.

## Tasks
- [x] Make the meal clickable on the recap screen
- [x] Change the `navigateToHome` on `addMealFlow` to be more generic, so that it can also navigate back to the daily recap
- [x] Make navigating to the daily recap clear the navigation stack of add meal, so that the back button doesn't improperly return there
- [x] Update data on daily recap after returning to the screen

##  Notes
Small refactor of the UI to include early returns to make things less cognitively complex.


app/src/androidTest/java/com/github/se/polyfit/ui/screen/DailyRecapScreens.kt
medwardson, line 1: The rename fixes JaCoCo line coverage issue......

app/src/main/java/com/github/se/polyfit/MainActivity.kt
```diff
@@ -57,9 +57,10 @@ class MainActivity : ComponentActivity() {
           composable(Route.Register) { LoginScreen(navigation::navigateToHome) }
           composable(Route.AddMeal + "/{mId}") { backStackEntry ->
             val mealId = backStackEntry.arguments?.getString("mId")
+            // If we allow edits from other screens, we will have to modify how we choose goForward
```
aprilslis, line 60: maybe we can provide another parameter? something like /home if you want to navigate back home and /recap if you want that? But we could leave that until its needed

app/src/main/java/com/github/se/polyfit/ui/screen/DailyRecapScreen.kt
```diff
@@ -36,33 +41,49 @@ fun DailyRecapScreen(
...
+    onDispose { lifecycleOwner.lifecycle.removeObserver(observer) }
+  }
 
   Scaffold(topBar = { SimpleTopBar("Overview", navigateBack) }) {
     Column(modifier = Modifier.padding(it).testTag("DailyRecapScreen")) {
       DateSelector(onConfirm = dailyRecapViewModel::getMealsOnDate, title = "Date")
       if (isFetching) {
         CircularProgressIndicator(
             modifier = Modifier.fillMaxSize().padding(16.dp).testTag("Spinner"))
+        return@Scaffold
```
aprilslis, line 65: Good changes! Maybe it could have been done in a separated PR

aprilslis (approved):
Great work, nothing major to say!

RandomUsername1315:
Just a heads up that I started to review your code and am not done yet

app/src/main/java/com/github/se/polyfit/viewmodel/dailyRecap/DailyRecapViewModel.kt
```diff
@@ -23,11 +26,15 @@ class DailyRecapViewModel @Inject constructor(private val mealRepository: MealRe
...
+
   init {
-    getMealsOnDate(LocalDate.now())
+    getMealsOnDate()
   }
 
-  fun getMealsOnDate(date: LocalDate) {
+  fun getMealsOnDate(newDate: LocalDate = date) {
+    date = newDate
+
```
antomax33, line 37: What is the purpose to put the old date as by default argument to the new date?

medwardson: So that it saves the date state upon returning from edit, otherwise we can not keep them on the same date when they return from edits

medwardson:
Update on message from @RandomUsername1315, good to merge it now and if there's anything pressing from his review then it will be addressed in a follow-up

PR #268, opened 2024-05-14
Get the Location to store a post

Neb1010:
### News
- we can now have the current location when we store a post.
- when we click on allow google will ask what kind of permissions you allow and if you accept automaticaly fetch the location and push the post into firebase


app/src/main/java/com/github/se/polyfit/model/post/MapLocationModel.kt
```diff
@@ -0,0 +1,124 @@
+/*
```
antomax33, line 1: You can delete this file

app/src/main/java/com/github/se/polyfit/model/post/PostLocationModel.kt
```diff
@@ -0,0 +1,63 @@
...
+import kotlinx.coroutines.tasks.await
+
+class PostLocationModel(private val context: Context) {
+
+  private val currentLocationRequest =
+      CurrentLocationRequest.Builder().setPriority(Priority.PRIORITY_HIGH_ACCURACY).build()
+
+  suspend fun getCurrentLocation(): com.github.se.polyfit.model.post.Location {
+    return checkLocationPermissions()
+  }
```
antomax33, line 23: You can import Location and directly use instead of calling com.github...
or an alias if another already got this name

app/src/main/java/com/github/se/polyfit/model/post/PostLocationModel.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    if (ActivityCompat.checkSelfPermission(
+        this.context, Manifest.permission.ACCESS_COARSE_LOCATION) !=
+        PackageManager.PERMISSION_GRANTED ||
+        ActivityCompat.checkSelfPermission(
+            this.context, Manifest.permission.ACCESS_FINE_LOCATION) !=
+            PackageManager.PERMISSION_GRANTED) {
+      Log.e("LocationPermissionsModel", "Location permissions not granted")
+
+      return com.github.se.polyfit.model.post.Location.default()
```
antomax33, line 58: Same for the import

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
```diff
@@ -60,8 +63,18 @@ constructor(
     _post.createdAt = createdAt
   }
 
-  fun setPostLocation(location: Location) {
-    _post.location = location
+  fun setPostLocation(): Job {
```
antomax33, line 66: It is unusual to have a setter without argument, maybe updatePostLocation or initPostLocation

antomax33 (approved):
No real problem, just some small changes to improve readability.

app/src/main/java/com/github/se/polyfit/model/post/PostLocationModel.kt
Neb1010: The problem: we have a `Location` and what it returns is a `Location` too, so I got some error because the compiler didn't know what Location I was meaning. But it should exist a real solution to this problem so if you know it, i'm all ear

app/src/main/java/com/github/se/polyfit/model/post/MapLocationModel.kt
Neb1010: Yes mb on this file, but I need to work on it later for the location for the map this time, because this is different and it will be the `model` for the location map, and I've already written some part of code that I will need and still work on this branch after the merge, if it's okay to let it

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
Neb1010: That's true, I just didn't change the current name when I modified it, but I can refactor in initPostLocation

app/src/main/java/com/github/se/polyfit/model/post/MapLocationModel.kt
antomax33: :+1:

app/src/main/java/com/github/se/polyfit/model/post/PostLocationModel.kt
antomax33: With android studio you can import by alias. So the name will be for example location2. So just import Location as location2.
import com.sorandom.app.model.api.Input as OtherInputfun map(input: Input): OtherInput {
  return OtherInput(
    ...
  )
}
https://medium.com/@erickpranata/kotlin-import-alias-3f2831295902

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
```diff
@@ -60,8 +63,18 @@ constructor(
...
+    return viewModelScope.launch(Dispatchers.IO) {
+      val location = postLocationModel.getCurrentLocation()
+      _post.location = location
+    }
+  }
+
+  fun setInOrder(first: Job = setPostLocation(), second: () -> Unit = ::setPost) {
```
RandomUsername1315, line 73: Maybe it would be better to remove the arguments and directly call the functions in the body of this function, and give it a more suitable name.

RandomUsername1315 (approved):
Nothing major just a few things here and there.

app/src/main/java/com/github/se/polyfit/ui/components/dialog/LocationPermissionDialog.kt
```diff
@@ -20,6 +25,11 @@ import com.github.se.polyfit.ui.theme.PrimaryPurple
 
 @Composable
 fun LocationPermissionDialog(onDeny: () -> Unit, onApprove: () -> Unit) {
+
+  val permissionLauncher = launcherForActivityResult(onDeny, onApprove)
```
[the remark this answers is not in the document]
Neb1010: I can do it like this if you provide the test for this part, just kidding, it's just for line coverage purpose otherwise i would have done like you

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
Neb1010: same It's just for testing the function,otherwise I don't manage to catch the execution of the function in the tests

app/src/main/java/com/github/se/polyfit/ui/components/dialog/LocationPermissionDialog.kt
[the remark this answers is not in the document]
RandomUsername1315: I see ok

app/src/main/java/com/github/se/polyfit/viewmodel/post/CreatePostViewModel.kt
RandomUsername1315: Do you mean that if you have a coroutine inside a ViewModel scope, you weren't able to catch the error when mocking it?

PR #270, opened 2024-05-15
Fixing things with the graphs

sharif-bbd:
## Fixing bugs and Adding to the Graph User Story

- [x] Making it clear that to access the graph screen you need to click the title of the card
- [x] Adding ascending and descending sorting
- [x] removing elemenst with calories count to 0 from being displayed in the list of elements while still being on graph


app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
Neb1010, line 1: Your `FullGraphScreen` is realy difficult to read and so to review (and probably to test)
It would be easier if you have modularized it, I saw you make multiple times a `HorizontalDivider` it would have been nice to make a seperate function among other things. It can be easier to spot and correct bugs in the future if you make more small `@composable`

Neb1010 (approved):
Your viewModel looks good, but your `@composable` are realy long, it will become difficult to maintain the code in the futur

antomax33 (approved):
LGTM

app/src/main/java/com/github/se/polyfit/ui/screen/GraphScreen.kt
sharif-bbd: Alright I will try to modularize it

app/src/main/java/com/github/se/polyfit/ui/components/Graphs.kt
```diff
@@ -26,6 +26,7 @@ fun lineChartData(
     dateList: List<LocalDate>,
     screen: DisplayScreen
 ): LineChartData {
+  val pointListCopy = pointList
```
aprilslis, line 29: im confused what this does? it seems like you could just use pointList?

PR #273, opened 2024-05-16
Account Settings

aprilslis:
UI, Viewmodel, tests for Account Settings. It is fully integrated with backend.

On this screen, Users can edit the information we have about them. The name fields are mandatory, and the rest is optional (since this stuff needs to be edited by the user because we don't get it from firebase). It is integrated with the bottom navigation bar and the main setting screen.

I haven't yet added the UI/sliders for the dietary restrictions because of time constraint, but I think this will be sufficient for M3 and then will add those in M4. The backend is setup for them already

Screenshot 2024-05-16 at 1 18 58 PM

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/MainActivity.kt
+++ b/app/src/main/java/com/github/se/polyfit/MainActivity.kt
@@ -4,6 +4,8 @@ import android.app.Application
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
+import androidx.compose.foundation.layout.padding
+import androidx.compose.ui.Modifier
 import androidx.core.view.WindowCompat
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsControllerCompat
@@ -74,7 +76,11 @@ class MainActivity : ComponentActivity() {
           }
 
           composable(Route.Settings) {
-            GenericScreen(navController = navController, content = { SettingFlow() })
+            GenericScreen(
+                navController = navController,
+                content = { paddingValues ->
+                  SettingFlow(modifier = Modifier.padding(paddingValues))
+                })
           }
 
           composable(Route.PostInfo) { PostInfoScreen(navigation, navController) }
--- a/app/src/main/java/com/github/se/polyfit/MainActivity.kt
+++ b/app/src/main/java/com/github/se/polyfit/MainActivity.kt
@@ -12,6 +12,7 @@ import androidx.navigation.compose.composable
 import androidx.navigation.compose.rememberNavController
 import com.github.se.polyfit.ui.components.GenericScreen
 import com.github.se.polyfit.ui.flow.AddMealFlow
+import com.github.se.polyfit.ui.flow.SettingFlow
 import com.github.se.polyfit.ui.navigation.Navigation
 import com.github.se.polyfit.ui.navigation.Route
 import com.github.se.polyfit.ui.screen.CreatePostScreen
@@ -70,6 +71,10 @@ class MainActivity : ComponentActivity() {
                 goBack = navigation::goBack,
                 goForward = { navigation.goBackTo(Route.DailyRecap) },
                 mealId = mealId)
+          }
+
+          composable(Route.Settings) {
+            GenericScreen(navController = navController, content = { SettingFlow() })
           }
 
           composable(Route.PostInfo) { PostInfoScreen(navigation, navController) }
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/github/se/polyfit/model/user/User.kt
```diff
@@ -2,30 +2,34 @@ package com.github.se.polyfit.model.data
...
-    photoURL = null
+    update(User()) // Creates an empty user and then updates the current user with it
   }
 
   fun isSignedIn(): Boolean {
     return id.isNotEmpty()
   }
 
-  fun serialize(): Map<String, Any> {
+  fun serialize(): Map<String, Any?> {
```
aprilslis, line 32: Since not everything has a value, if it's null it doesn't get stored in firebase and then will get reinitialized as null upon deserialiazation through `as?` method

app/src/main/java/com/github/se/polyfit/model/user/User.kt
```diff
@@ -45,37 +57,73 @@ data class User(
...
+              "familyName" to user.familyName!!,
+              "givenName" to user.givenName!!,
+              "email" to user.email,
+              "calorieGoal" to user.calorieGoal,
+              "isVegan" to user.isVegan,
+              "isVegetarian" to user.isVegetarian,
+              "isGlutenFree" to user.isGlutenFree,
+              "isLactoseFree" to user.isLactoseFree)
+
+      user.photoURL?.let { result["photoURL"] = it.toString() }
```
aprilslis, line 102: For optional ones, only include in serializations if its definted

app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
aprilslis, line 1: Sorry for the long file, there's not really much way to break it down, as even though there's lots of text fields they often don't share the same exact setup

app/src/main/java/com/github/se/polyfit/MainActivity.kt
```diff
@@ -64,6 +67,14 @@ class MainActivity : ComponentActivity() {
                 mealId = mealId)
           }
 
+          composable(Route.Settings) {
+            GenericScreen(
+                navController = navController,
```
Neb1010, line 72:
```
                navController = navController,
```
:(

app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
Neb1010: ### Improvement 
- The logic for TextField components can be extracted into a reusable function to reduce redundancy

That's the only point I notice.

Good Job!

app/src/main/java/com/github/se/polyfit/ui/flow/SettingFlow.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.navigation.compose.rememberNavController
+import com.github.se.polyfit.model.settings.SettingOption
+import com.github.se.polyfit.ui.navigation.Navigation
+import com.github.se.polyfit.ui.navigation.Route
+import com.github.se.polyfit.ui.screen.SettingsScreen
+import com.github.se.polyfit.ui.screen.settings.AccountSettingsScreen
+
+@Composable
+fun SettingFlow(modifier: Modifier = Modifier) {
+  val navController = rememberNavController()
```
Neb1010, line 18: :( 
```
```

Neb1010 (approved):
Good Job for your hard work

app/src/main/java/com/github/se/polyfit/MainActivity.kt
aprilslis: Yea, I know :( I followed the same format as the other files in this file. We can revamp later in its separate PR :D

app/src/main/java/com/github/se/polyfit/ui/flow/SettingFlow.kt
aprilslis: This one is ok :P This flow has it's own nav controller, it needs to be defined here

app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
aprilslis: It can be, but it really doesn't save too much space. But, I guess it makes it easy to change things for the future. Thanks for the suggestion

PR #283, opened 2024-05-17
Prevent PrimaryButton double clicks

medwardson:
Only run the onClick of the primary button once, to prevent unexpected side-effects of spam

### DIFF
--- a/app/src/androidTest/java/com/github/se/polyfit/ui/components/button/PrimaryButtonTest.kt
+++ b/app/src/androidTest/java/com/github/se/polyfit/ui/components/button/PrimaryButtonTest.kt
@@ -11,14 +11,14 @@ import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.se.polyfit.ui.components.GradientBox
-import com.github.se.polyfit.ui.navigation.Navigation
 import com.kaspersky.kaspresso.testcases.api.testcase.TestCase
 import io.github.kakaocup.compose.node.element.ComposeScreen
-import io.mockk.impl.annotations.RelaxedMockK
 import io.mockk.junit4.MockKRule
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/polyfit/ui/components/button/PrimaryButton.kt
```diff
@@ -24,8 +28,15 @@ fun PrimaryButton(
...
 ) {
+  var pressed by remember { mutableStateOf(false) }
+
   Button(
-      onClick = onClick,
+      onClick = {
+        if (!pressed) {
+          onClick()
+          pressed = true
+        }
```
antomax33, line 38: Looks good, maybe pressed = true before onClick() but not important

app/src/androidTest/java/com/github/se/polyfit/ui/components/button/PrimaryButtonTest.kt
```diff
@@ -79,4 +77,24 @@ class PrimaryButtonTest : TestCase() {
...
+    }
+    ComposeScreen.onComposeScreen<GradientBox>(composeTestRule) {
+      composeTestRule.onNodeWithTag("PrimaryButton").assertIsDisplayed()
+      composeTestRule.onNodeWithTag("PrimaryButton").assertHasClickAction()
+      composeTestRule.onNodeWithTag("PrimaryButton").performClick()
+      composeTestRule.onNodeWithTag("PrimaryButton").performClick()
+    }
+
+    verify(exactly = 1) { onClick() }
+  }
```
antomax33, line 99: Great to add a test every time a bug is fixed.

antomax33 (approved):
LGTM

app/src/main/java/com/github/se/polyfit/ui/components/button/PrimaryButton.kt
medwardson: Nice idea, fixed

PR #284, opened 2024-05-17
Ingredient button disables until ingredient is added

medwardson:
User shouldn't be able to go to next step of meal creation without adding ingredients


aprilslis (approved):
Nice job and good job cleaning up the dead code

PR #285, opened 2024-05-17
Fix issues with editing ingredient info

aprilslis:
Now, when you type `1` it will say `1.0` instead of `10.0`. Also, as soon as you type it, the button will activate.


medwardson (approved):
Awesome job, this makes the nutrient writing so much smoother

PR #288, opened 2024-05-18
Show a helpful message when posting if there are no meals

aprilslis:
When making a post, if you have no meals it will bring you to the create post screen and you won't be able to select anything. Instead, let's show a helpful message that directs users to add a meal. Also small fix is to round the calories to two decimals to fix formatting display issue

image

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/ui/components/selector/MealSelector.kt
+++ b/app/src/main/java/com/github/se/polyfit/ui/components/selector/MealSelector.kt
@@ -82,7 +82,10 @@ fun MealSelector(
                   modifier = Modifier.weight(1f),
               )
               Text(
-                  text = "${selectedMeal.calculateTotalCalories().roundTwoDecimal()} kcal",
+                  text =
+                      context.getString(
+                          R.string.kcalvalue,
+                          selectedMeal.calculateTotalCalories().roundTwoDecimal()),
                   fontSize = 20.sp,
                   modifier = Modifier.weight(1f).padding(end = 20.dp),
                   textAlign = TextAlign.End,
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/polyfit/ui/components/selector/MealSelector.kt
```diff
@@ -82,7 +82,7 @@ fun MealSelector(
                   modifier = Modifier.weight(1f),
               )
               Text(
-                  text = "${selectedMeal.calculateTotalCalories()} kcal",
+                  text = "${selectedMeal.calculateTotalCalories().roundTwoDecimal()} kcal",
```
sharif-bbd, line 85: it could be nice to use the String ressource file to pass your strings so something like this

```
context.getString(R.string.calories, selectedMeal.calculateTotalCalories().roundTwoDecimal())
```
and then in the string you could put something like 
```
    <string name="calories">%1$.2f kCal</string>
```

app/src/main/java/com/github/se/polyfit/ui/screen/CreatePostScreen.kt
```diff
@@ -78,6 +81,11 @@ fun CreatePostScreen(
         BottomBar(
             postComplete = postComplete, onButtonPressed = { isPermissionDialogDisplay = true })
       }) {
+        if (meals.isEmpty()) {
+          CreateMeal(navigateToAddMeal, Modifier.padding(it))
+          return@Scaffold
```
sharif-bbd, line 86: that's good practice !

sharif-bbd (approved):
Great Idea for the message ! good job

PR #289, opened 2024-05-18
OverviewScreen daily calories not being displayed correctly

RandomUsername1315:
## Bugfix
The values for the daily summary of calories consumed for each meal were hardcoded instead of being fetched from the local database. This issue also affected the calorie goals.

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
+++ b/app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
@@ -1,5 +1,7 @@
 package com.github.se.polyfit.data.processor
 
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.map
 import com.github.se.polyfit.data.local.dao.MealDao
 import com.github.se.polyfit.model.meal.MealOccasion
 import java.time.LocalDate
@@ -19,6 +21,26 @@ class LocalDataProcessor @Inject constructor(private val mealDao: MealDao) {
 
   fun getCaloriesPerMealOccasionToday(): Map<MealOccasion, Double> {
     return getCaloriesPerMealOnSpecificDay(LocalDate.now())
+  }
+
+  fun getCaloriesPerMealOccasionTodayLiveData(): LiveData<List<Pair<MealOccasion, Double>>> {
+    return mealDao.getMealsCreatedOnOrAfterDateLiveData(LocalDate.now()).map { meals ->
+      val allOccasion =
+          MealOccasion.values().toList().filter { it != MealOccasion.OTHER }.map { Pair(it, 0.0) }
+
+      val mealValues =
+          meals
+              .toMutableList()
+              .groupBy { it.occasion }
+              .mapValues { (_, meals) -> meals.sumOf { it.calculateTotalCalories() } }
+              .toList()
+
+      allOccasion.map {
+        val (occasion, _) = it
+        val value = mealValues.find { it.first == occasion }?.second ?: 0.0
+        Pair(occasion, value)
+      }
+    }
   }
 
   fun getCaloriesPerMealOnSpecificDay(day: LocalDate): Map<MealOccasion, Double> {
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
```diff
@@ -26,6 +28,26 @@ class LocalDataProcessor @Inject constructor(private val mealDao: MealDao) {
     return getCaloriesPerMealOnSpecificDay(LocalDate.now())
   }
 
+  fun getCaloriesPerMealOccasionTodayLiveData(): LiveData<List<Pair<MealOccasion, Double>>> {
+    return mealDao.getMealsCreatedOnOrAfterDateLiveData(LocalDate.now()).map { meals ->
+      val allOccasion =
+          MealOccasion.values().toList().filter { it != MealOccasion.OTHER }.map { Pair(it, 0.0) }
```
sharif-bbd, line 34: shouldn't you also make sure the meal is complete ?

app/src/main/java/com/github/se/polyfit/ui/screen/OverviewScreen.kt
```diff
@@ -72,6 +73,12 @@ fun OverviewScreen(
   val iconExample = BitmapFactory.decodeResource(context.resources, R.drawable.picture_example)
   var imageBitmap by remember { mutableStateOf(iconExample) }
 
+  var mealSummary by remember { mutableStateOf(listOf<Pair<MealOccasion, Double>>()) }
```
sharif-bbd, line 76: I think It would be good to create a LiveData variable in the viewModel to observe as state in the overviewScreen

app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
```diff
@@ -208,4 +211,13 @@ constructor(
       }
     }
   }
+
+  fun getCaloriesPerMealOccasionTodayLiveData(): LiveData<List<Pair<MealOccasion, Double>>> {
+
+    return localDataProcessor.getCaloriesPerMealOccasionTodayLiveData()
+  }
+
```
sharif-bbd, line 219: so you could do something like
```
private val _caloriesPerMealOccasions = MediatorLiveData<List<Pair<MealOccasion, Double>>>
// This creates a readOnly LiveData variable
private val caloriesPerMealOccasions : LiveData<List<Pair<MealOccasion, Double>>> = caloriesPerMealOccasions

```

And then set the value of `caloriesPerMealOccasions` to `getCaloriesPerMealOccasionTodayLiveData()`

either in the viewModel or with something like this in the overviewScreen
```
mealSummary by viewModel.caloriesPerMealOccasions.observeAsStata(viewModel.getCaloriesPerMealOccasionTodayLiveData())
```

To keep the separation between the viewModel and the screen

sharif-bbd (approved):
Good initiative to solve this issue. Excellent !

app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
RandomUsername1315: Yep catch. This will avoid displaying uncompleted meal !

app/src/main/java/com/github/se/polyfit/ui/screen/OverviewScreen.kt
RandomUsername1315: That is a great idea and is already happening. On line 78:

```kotlin
  LaunchedEffect(Unit) {
    overviewViewModel.getCaloriesPerMealOccasionTodayLiveData().observeForever { mealSummary = it }
  }
```

An observer is set the first time the screen is displayed on the LiveData. The LiveData is not created at the level of the overviewViewModel but at the level of the database.

app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
RandomUsername1315: That is a great idea and that is basically what is implemented. The live data created in the local database and then it is passed to the overview screen where it is observed.

PR #290, opened 2024-05-18
Use lazy column for settings so it doesn't get blocked

aprilslis:
So that the keyboard doesn't cover the bottom of the settings, make it a lazy column. Tested with a smaller sized emulator:

Screenshot 2024-05-18 at 2 39 40 PM


RandomUsername1315 (approved):
LGTM. I tested it on my phone and it fixes the issue !

PR #291, opened 2024-05-18
Import picture button not behaving correctly

RandomUsername1315:
## Bugfix
When trying to create a meal using Spoonacular, the user is given two choices: either take a picture and use that, or choose an existing picture from their library. The button to import a picture existed but did not do anything.

---
### Small Note
The launcher used by the user to import picture cannot be tested (sadly). In order to merge, the coverage check for new code coverage will have to be disabled. 

(edit : with the help of @antomax33 I was able to test the thing with the intent)


medwardson:
@RandomUsername1315 before testing & reviewing, is it ok to do so right now since it'll use spoonacular calls?

RandomUsername1315:
@medwardson yeah it's as long as you don't make like 50 calls it should be ok.

app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
```diff
@@ -220,4 +224,29 @@ constructor(
...
+  ): String? {
+    if (uri != null) {
+      val imageBitmap =
+          uriToBitmap(context, uri)
+              ?: BitmapFactory.decodeResource(context.resources, R.drawable.picture_example)
+      return overviewViewModel.storeMeal(imageBitmap)
+    } else {
+      Log.d("PhotoPicker", "No media selected")
+      return null
+    }
```
medwardson, line 250:
```suggestion
     return if (uri != null) {
      val imageBitmap =
        uriToBitmap(context, uri)
          ?: BitmapFactory.decodeResource(context.resources, R.drawable.picture_example)
      overviewViewModel.storeMeal(imageBitmap)
    } else {
      Log.d("PhotoPicker", "No media selected")
      null
    }
```
Simplification that android studio suggests

medwardson (commented):
Looks good Charles. The only concern I have is that when I go to upload and then cancel the upload (as in, don't choose a picture and close the drawer), it will still take me to the add meal flow and not have any meal loaded. I think it would be better to just keep on the main screen in this case. Thoughts?

RandomUsername1315:
@medwardson Yeah, I didn't notice this when testing. I mean, the behavior mirrors the food recognition. But yeah, I think you're completely right—it improves the user experience if they are not automatically taken to the create ingredient page. I will try to fix that.

app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
RandomUsername1315: Nice suggestion !

RandomUsername1315:
@medwardson what do you think of this : 

https://github.com/swent-group10/polyfit/assets/100274930/df5d15ce-7d83-4ebd-b700-65095e6c4da4

medwardson:
Looks awesome!

RandomUsername1315:
@medwardson yeah I forgot to kftm that is why 100 lines of codes turned into 1000

medwardson:
It looks like some of the changes from recipe recommendation flow ended up in here

RandomUsername1315:
@medwardson yep thanks for catching that, some random changes from another pr slipped in.

app/build.gradle.kts
```diff
@@ -276,3 +276,6 @@ tasks.register("jacocoTestReport", JacocoReport::class) {
         include("outputs/code_coverage/debugAndroidTest/connected/*/coverage.ec")
     })
 }
+dependencies {
+    implementation("androidx.navigation:navigation-testing:2.7.7")
+}
```
antomax33, line 281: Not here the dependencies

app/src/main/java/com/github/se/polyfit/ui/screen/OverviewScreen.kt
```diff
@@ -134,14 +152,11 @@ fun OverviewScreen(
...
-                          context, context.getString(R.string.PermissionDenied), Toast.LENGTH_SHORT)
-                      .show()
+                  pickMedia.launch(
+                      PickVisualMediaRequest(ActivityResultContracts.PickVisualMedia.ImageOnly))
                 },
                 {
-                  Toast.makeText(
-                          context, context.getString(R.string.PermissionDenied), Toast.LENGTH_SHORT)
-                      .show()
+                  Toast.makeText(context, "Feature not yet implemented", Toast.LENGTH_SHORT).show()
```
antomax33, line 159: The old one is better, all the shown text is in the string.xml file

app/src/main/java/com/github/se/polyfit/ui/navigation/Navigation.kt
```diff
@@ -72,4 +72,8 @@ class Navigation(private val navHostController: NavHostController) {
     Log.i("Navigation", "Navigating to $route")
     navHostController.navigate(route)
   }
+
+  fun navigateToPostInfo() {
+    navigateTo(Route.PostInfo)
+  }
```
antomax33, line 82: I'm not sure, but is it better than just calling navigateTo(Route.PostInfo)?

app/build.gradle.kts
RandomUsername1315: Yeah, automatically importing the dependencies does that for some reason, and I always move it. Thanks for catching it.

app/src/main/java/com/github/se/polyfit/ui/screen/OverviewScreen.kt
RandomUsername1315: How is the message @antomax33  ? The text will be moved to the string.xml`

app/src/main/java/com/github/se/polyfit/ui/navigation/Navigation.kt
RandomUsername1315: I just tried to align myself with the practices of the overall codebase. I think that it looks a bit cleaner because there is no need to pass the destination as an argument.

app/src/main/res/values/strings.xml
```diff
@@ -62,10 +62,13 @@
...
     <string name="addMealButton">Add a Meal</string>
+    <string name="ErrorPicture">Error fetching the picture</string>
     <string name="MoreDetailRecipe">More Detail</string>
     <string name="Ingredients">Ingredients</string>
     <string name="Recipe">Recipe</string>
 
     <string name="confirmSignout">Are you sure you want to sign out?</string>
     <string name="signout">Sign Out</string>
+    <string name="FeatureNotYetImplemented">Feature not yet implemented</string>
+    Feature not yet implemented
```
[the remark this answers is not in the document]
RandomUsername1315: Nice catch !

app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
```diff
@@ -220,4 +224,29 @@ constructor(
...
+
+  fun handleSelectedImage(
+      context: Context,
+      uri: Uri?,
+      overviewViewModel: OverviewViewModel,
+  ): String? {
+    return if (uri != null) {
+      val imageBitmap =
+          uriToBitmap(context, uri)
+              ?: BitmapFactory.decodeResource(context.resources, R.drawable.picture_example)
```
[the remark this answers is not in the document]
RandomUsername1315: Yeah you're right. That a really fun bug to find.

app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
```diff
@@ -220,4 +224,29 @@ constructor(
...
     return user.calorieGoal
   }
+
+  fun uriToBitmap(context: Context, uri: Uri): Bitmap? {
+    return context.contentResolver
+        .openInputStream(uri)
+        ?.use { inputStream ->
+          ByteArrayOutputStream().apply { inputStream.copyTo(this) }.toByteArray()
+        }
+        ?.let { BitmapFactory.decodeByteArray(it, 0, it.size) }
```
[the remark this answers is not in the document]
RandomUsername1315: Completely right. I simplified the code and added more comments.

app/src/main/java/com/github/se/polyfit/ui/screen/OverviewScreen.kt
```diff
@@ -231,3 +250,11 @@ fun OverviewScreen(
...
   }
 }
+
+fun navigateOrShowError(id: String?, navigation: Navigation, context: Context) {
+  if (id != null) {
+    navigation.navigateToAddMeal(id)
+  } else {
+    Toast.makeText(context, context.getString(R.string.ErrorPicture), Toast.LENGTH_SHORT).show()
+  }
+}
```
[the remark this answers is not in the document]
RandomUsername1315: That would be a good idea, but the navigation class already handles the case where the ID is `null` differently. The current implementation creates a new meal if the ID is `null`.

PR #292, opened 2024-05-18
Issue with clicking twice on the back arrow

RandomUsername1315:
## Bugfix

A small issue occurred when the user clicked twice on the back arrow in some screens before the screen could be updated with the new UI. This was caused by the `NavHostController` being `popBackStack()` twice instead of once. This is resolved by allowing only the first click to be registered before navigating to the new screen.

### DIFF
--- a/app/src/androidTest/java/com/github/se/polyfit/ui/screen/CreatePostTest.kt
+++ b/app/src/androidTest/java/com/github/se/polyfit/ui/screen/CreatePostTest.kt
@@ -155,11 +155,17 @@ class CreatePostTest : TestCase() {
         assertExists()
         assertIsDisplayed()
         assertIsEnabled()
-      }
-    }
 
-    ComposeScreen.onComposeScreen<CreatePostBottomBar>(composeTestRule) {
-      composeTestRule.onNodeWithTag("BackButton").assertExists().assertIsDisplayed().performClick()
+        composeTestRule
+            .onNodeWithTag("BackButton")
+            .assertExists()
+            .assertIsDisplayed()
+            .performClick()
+            .performClick()
+
+        // count the number of times a function is called mockNavForward
+        verify(exactly = 1) { mockNavBack() }
+      }
     }
   }
 }
--- a/app/src/androidTest/java/com/github/se/polyfit/ui/screen/CreatePostTest.kt
+++ b/app/src/androidTest/java/com/github/se/polyfit/ui/screen/CreatePostTest.kt
@@ -1,7 +1,10 @@
 package com.github.se.polyfit.ui.screen
 
 import android.graphics.Bitmap
+import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performClick
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.se.polyfit.data.remote.firebase.PostFirebaseRepository
 import com.github.se.polyfit.data.repository.MealRepository
@@ -153,6 +156,10 @@ class CreatePostTest : TestCase() {
         assertIsDisplayed()
         assertIsEnabled()
       }
+    }
+
+    ComposeScreen.onComposeScreen<CreatePostBottomBar>(composeTestRule) {
+      composeTestRule.onNodeWithTag("BackButton").assertExists().assertIsDisplayed().performClick()
     }
   }
 }
--- a/app/src/main/java/com/github/se/polyfit/MainActivity.kt
+++ b/app/src/main/java/com/github/se/polyfit/MainActivity.kt
@@ -86,7 +86,8 @@ class MainActivity : ComponentActivity() {
           composable(Route.PostInfo) { PostInfoScreen(navigation, navController) }
 
           composable(Route.CreatePost) {
-            CreatePostScreen(navigation::goBack, navigation::navigateToPostList)
+            CreatePostScreen(
+                navigation::goBack, navigation::navigateToPostList, navigation::navigateToAddMeal)
           }
 
           composable(Route.DailyRecap) {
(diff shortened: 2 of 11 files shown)


app/src/androidTest/java/com/github/se/polyfit/ui/screen/CreatePostTest.kt
```diff
@@ -154,5 +157,9 @@ class CreatePostTest : TestCase() {
         assertIsEnabled()
       }
     }
+
+    ComposeScreen.onComposeScreen<CreatePostBottomBar>(composeTestRule) {
+      composeTestRule.onNodeWithTag("BackButton").assertExists().assertIsDisplayed().performClick()
+    }
```
medwardson, line 163: This could be included in the `ComposeScreen.onComposeScreen<CreatePostBottomBar>(composeTestRule) {` above on line 153

app/src/main/java/com/github/se/polyfit/MainActivity.kt
```diff
@@ -86,8 +86,7 @@ class MainActivity : ComponentActivity() {
           composable(Route.PostInfo) { PostInfoScreen(navigation, navController) }
 
           composable(Route.CreatePost) {
-            CreatePostScreen(
-                navigation::goBack, navigation::navigateToHome, navigation::navigateToAddMeal)
+            CreatePostScreen(navigation::goBack, navigation::navigateToPostList)
```
medwardson, line 89: How come we're removing the navigateToAddMeal?

app/src/main/java/com/github/se/polyfit/ui/components/scaffold/CenteredTopBar.kt
```diff
@@ -28,7 +33,12 @@ fun CenteredTopBar(title: String, titleColor: Color = PrimaryPurple, navigateBac
       },
       navigationIcon = {
         IconButton(
-            onClick = { navigateBack() },
+            onClick = {
+              if (!isBackClicked) {
+                navigateBack()
+                isBackClicked = true
```
medwardson, line 39:
```suggestion
                isBackClicked = true
                navigateBack()
```
just as a safety to ensure its blocked before a second function runs

app/src/main/java/com/github/se/polyfit/ui/components/scaffold/SimpleTopBar.kt
```diff
@@ -28,7 +33,12 @@ fun SimpleTopBar(title: String, navigateBack: () -> Unit) {
       },
       navigationIcon = {
         IconButton(
-            onClick = { navigateBack() },
+            onClick = {
+              if (!goBackClicked) {
+                navigateBack()
+                goBackClicked = true
```
medwardson, line 39:
```suggestion
                goBackClicked = true
                navigateBack()
```

app/src/main/java/com/github/se/polyfit/ui/screen/IngredientScreen.kt
```diff
@@ -109,6 +111,7 @@ private fun BottomBar(onClickAddIngred: () -> Unit, navigateForward: () -> Unit,
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 private fun TopBar(navigateBack: () -> Unit) {
```
medwardson, line 113: maybe out of scope, but, it may be better to just replace this composable with the simpletopbar to improve future maintainability

app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
```diff
@@ -57,6 +60,7 @@ fun NutritionScreen(
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 private fun TopBar(navigateBack: () -> Unit) {
+  var goBackCliked by remember { mutableStateOf(false) }
```
medwardson, line 63: same

medwardson (approved):
Looks good! It may be worthwhile also adding a test in Simple/Centered top bar where you call `performClick` twice on the back button and verify the function is only called once

app/src/main/java/com/github/se/polyfit/ui/components/scaffold/CenteredTopBar.kt
RandomUsername1315: Ok thanks. I did not it could cause an issue. I will fix it right away !

app/src/main/java/com/github/se/polyfit/MainActivity.kt
RandomUsername1315: Looking at the arguments of the function:

```kotlin
fun CreatePostScreen(
    navigateBack: () -> Unit = {},
    navigateForward: () -> Unit = {},
    navigateToAddMeal: () -> Unit = {},
    postViewModel: CreatePostViewModel = hiltViewModel()
)

``` 
The previous implementation takes the user to the overview screen instead of the add meal flow for some reason when the meal is empty. However, because the button is always disabled when the meal is empty, that functionality does not work. I think it makes more sense to take the user to the list of posts after creating a post rather than to the overview.

app/src/main/java/com/github/se/polyfit/ui/screen/IngredientScreen.kt
RandomUsername1315: Yeah something I noticed while going through the code. I saw a `TODO` somewhere about that. I agree with  you I will add it to the scrum board.

app/src/main/java/com/github/se/polyfit/MainActivity.kt
RandomUsername1315: Yeah I added it back. But it was originally in the spot of navigateForwards which caused some weird behavior.

RandomUsername1315: I updated my main and I guess that that is gone so it was fixed by another PR.

PR #297, opened 2024-05-19
Logout

aprilslis:
Allow users to logout. Gives a popup to confirm if they really want to logout first

Screenshot 2024-05-19 at 11 16 07 AM

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/viewmodel/settings/SettingFlowViewModel.kt
+++ b/app/src/main/java/com/github/se/polyfit/viewmodel/settings/SettingFlowViewModel.kt
@@ -1,0 +1,17 @@
+package com.github.se.polyfit.viewmodel.settings
+
+import androidx.lifecycle.ViewModel
+import com.github.se.polyfit.model.data.User
+import com.google.firebase.auth.FirebaseAuth
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+
+@HiltViewModel
+class SettingFlowViewModel
+@Inject
+constructor(private val user: User, private val firebaseInstance: FirebaseAuth) : ViewModel() {
+  fun signOut() {
+    user.signOut()
+    firebaseInstance.signOut()
+  }
+}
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/se/polyfit/viewmodel/settings/SettingFlowViewModel.kt
```diff
@@ -0,0 +1,17 @@
...
+import javax.inject.Inject
+
+@HiltViewModel
+class SettingFlowViewModel
+@Inject
+constructor(private val user: User, private val firebaseInstance: FirebaseAuth) : ViewModel() {
+  fun signOut() {
+    user.signOut()
+    firebaseInstance.signOut()
+  }
```
RandomUsername1315, line 16: Something to keep in mind with the way everything is set up is that if a user logs out and logs in with another account, all their information will be stored in the first account. This is because neither the Firebase nor the local database is created when logging out.

RandomUsername1315 (approved):
Everything looks good. Just potential issue with storing data after signing out.

app/src/main/java/com/github/se/polyfit/viewmodel/settings/SettingFlowViewModel.kt
aprilslis: Oh im not quite sure what you mean? So we should clear the local database too when they logout?

medwardson: A possible suggestion is that we include user_id as a foreign key on the local meals db, this way we can query for meals that belong to a specific user only

RandomUsername1315: The issue with Hilt is that it initializes the arguments once. When they are created, it passes them the current user ID at that time. If the user logs out, the Firebase database will not be reinitialized to store the data in the new user's folder. For the local database, it will still contain the previous user's meals. So it might be a good idea to do as mesha suggested or once the user logs in to fetch all the meals from the firebase and update the local db with that.

PR #307, opened 2024-05-19
say third button disabled and rename to scan barcode

antomax33:


RandomUsername1315 (approved):
Thank you for fixing this bug

PR #315, opened 2024-05-21
Add userId to meals

medwardson:
Update the `Meal` class to have a `userId` to prevent issues when a different user logs in and our database stores multi-user info. This will allow us to call `getAllMeals()` and other meal queries with a injected parameters of userId, to only return meals relevant to the logged in user.

<hr>
I've ran a migration script on our Firebase user database to backfill all the meals to include the userId, and also for our posts. I'll run it again after merge. Here's the scripts I used:

$$\quad$$

```js
const admin = require('firebase-admin');
const serviceAccount = require('./myPrivateKey');  // downloaded from firebase

admin.initializeApp({
    credential: admin.credential.cert(serviceAccount),
    databaseURL: 'https://<your-database-name>.firebaseio.com' // copy real from firebase
});

const db = admin.firestore();

// Update Posts
async function updatePostsMealUserIdField() {
    const postsCollection = db.collection('posts');

    try {
        const postsSnapshot = await postsCollection.get();

        if (postsSnapshot.empty) {
            console.log('No posts found.');
            return;
        }

        const batch = db.batch();

        postsSnapshot.forEach(postDoc => {
            const postRef = postsCollection.doc(postDoc.id);
            const meal = postDoc.data().meal || {};
            meal.userId = ''; 
            batch.update(postRef, { meal: meal });
        });

        await batch.commit();
        console.log('Updated userId field in meal for all posts.');
    } catch (error) {
        console.error('Error updating userId field in meal for posts:', error);
    }
}

// Update user's meals
async function backfillUserId() {
    const usersCollection = db.collection('users');

    try {
        const usersSnapshot = await usersCollection.get();
        
        for (const userDoc of usersSnapshot.docs) {
            const userId = userDoc.id;
            const mealsCollection = usersCollection.doc(userId).collection('meals');
            const mealsSnapshot = await mealsCollection.get();

            const batch = db.batch();

            mealsSnapshot.forEach(mealDoc => {
                const mealRef = mealsCollection.doc(mealDoc.id);
                batch.update(mealRef, { userId: userId });
            });

            await batch.commit();
            console.log(`Backfilled userId for meals of user ${userId}`);
        }

        console.log('Backfill complete for all users.');
    } catch (error) {
        console.error('Error backfilling userId:', error);
    }
}
```

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
+++ b/app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
@@ -35,18 +35,11 @@ constructor(
     if (imageBitmap == null) {
       Log.e("OverviewViewModel", "Image is null")
       return null
-    } else {
-      val meal = spoonacularApiCaller.getMealsFromImage(imageBitmap)
-
-      val mealId = mealDao.insert(meal)
-
-      if (mealId == null) {
-        Log.e("OverviewViewModel", "Meal could not be inserted into the database")
-        return null
-      } else {
-        return mealId
-      }
     }
+
+    val meal = spoonacularApiCaller.getMealsFromImage(imageBitmap)
+    meal.userId = user.id
+    return mealDao.insert(meal)
   }
 
   fun deleteById(mealDatabaseId: String) {
@@ -55,7 +48,7 @@ constructor(
 
   fun getMealsByCalorieRange(minCalories: Double, maxCalories: Double): List<Meal> {
     // Get all meals from the database
-    val allMeals = mealDao.getAllMeals()
+    val allMeals = mealDao.getAllMeals(user.id)
 
     // Filter meals based on their calorie content
     val filteredMeals =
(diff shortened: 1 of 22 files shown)


RandomUsername1315:
@medwardson, it might be a good idea to wait before running the script to avoid any potential issues with the coaches' APK. I know that it shouldn't cause problems, but let's do that after the grading to avoid any potential crashes.

app/src/main/java/com/github/se/polyfit/viewmodel/meal/OverviewViewModel.kt
```diff
@@ -35,18 +35,11 @@ constructor(
     if (imageBitmap == null) {
       Log.e("OverviewViewModel", "Image is null")
       return null
-    } else {
-      val meal = spoonacularApiCaller.getMealsFromImage(imageBitmap)
-
-      val mealId = mealDao.insert(meal)
-
-      if (mealId == null) {
```
medwardson, line 43: It was impossible for mealId to be null, so simplified

antomax33 (approved):
Looks good to me

PR #320, opened 2024-05-25
Feature/backend/rec recipe

RandomUsername1315:
# Pull Request Overview

This PR adds the backend functionality for the recipe recommendation. It includes all new API calls necessary and the deserialization of the JSON.

## Main Changes

- **Create `RecipeInstructionResponseAPI.kt` class**: Represents the response of possible recipes that can be made given a list of ingredients from Spoonacular.
- **Modify `SpoonacularAPI.kt` class**: Added two functions, one that fetches the steps/instructions to make a recipe and another that handles all the calls necessary to create complete recipes with steps (two API calls) given a list of strings.
- **Create `RecipeInstructions.kt` class**: A card that displays a single step for a recipe.
- **Modify `RecipeRecommendationMoreDetailScreen.kt` class**: Shows a loading animation while it fetches data from the API.

## Video of the New Implementation
Just for reference what you see in the video made **12 API calls** :

https://github.com/swent-group10/polyfit/assets/100274930/cd1d6dbc-7074-4f8c-9455-e0ba057c48ea

## Future PR
- Integrating the barcode and the recipe recommendation to create a coherent feature.

## Note for Reviewers
- A few things had to be removed due to constraints from the API. For example, the calories and nutritional facts were removed because it would require making at least two more calls per recipe and creating new deserialization and data classes.
- Also, due to the high number of API calls made by this class, the first and most expensive call that returns a list of possible recipes given a list of strings has been **hardcoded** to return only one value for now. The other calls are made as they normally would be, so please avoid making too many calls.
- Also waiting on the fix from #321 to merge

# Note 

- **The PR is about 500 lines, I also refactored JSON variables used in tests to be in their own folder to make things more readable causing the addition metric to be this high**


app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCallerTest.kt
```diff
@@ -282,11 +294,114 @@ class SpoonacularApiCallerTest {
...
+    assertEquals(1, response.steps[0].number)
+  }
+
+  @Test
+  fun recipeStepFromIdFailure() {
+    mockWebServer.dispatcher = faultyDispatcher
+
+    val response = spoonacularApiCaller.getRecipeSteps(1)
+
+    assertEquals(response, RecipeInstruction.faillure())
```
medwardson, line 312:
```suggestion
    assertEquals(response, RecipeInstruction.failure())
```

app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCallerTest.kt
```diff
@@ -282,11 +294,114 @@ class SpoonacularApiCallerTest {
...
+    assert(response.first().recipeInformation.ingredients.isEmpty())
+    assert(response.first().recipeInformation.instructions.isEmpty())
+  }
+
   @After
   fun tearDown() {
     mockWebServer.shutdown()
   }
 
+  private val recipeStepsMock =
```
medwardson, line 339: id maybe suggest to create a folder in the test files called mockData and include this and large json test data there to keep the test file more logic focused

app/src/androidTest/java/com/github/se/polyfit/model/recipe/RecipeTest.kt
```diff
@@ -63,14 +66,16 @@ class RecipeTest {
     assertEquals(5L, result["usedIngredients"])
     assertEquals(2L, result["missingIngredients"])
     assertEquals(100L, result["likes"])
-    assertEquals("instructions", (result["recipeInformation"] as Map<String, Any>)["instructions"])
+    assertEquals(
+        listOf("instructions"), (result["recipeInformation"] as Map<String, Any>)["instructions"])
   }
 
   @Test
   fun desializeAndDeserializeDefaultMeal() {
```
medwardson, line 74:
``` suggestion
  fun serializeAndDeserializeDefaultMeal() {
```

app/src/androidTest/java/com/github/se/polyfit/ui/screen/recipeRec/RecipeRecommendationMoreDetailScreenTest.kt
medwardson, line 1: For file naming here, id suggest since the folder is already called recipeRec that this just becomes `MoreDetailScreen`, otherwise the name will become very long and the module it is in implies that it is recipe recommendation related

app/src/androidTest/java/com/github/se/polyfit/viewmodel/recipe/RecipeRecommendationViewModelTest.kt
```diff
@@ -30,12 +30,14 @@ class RecipeRecommendationViewModelTest {
     val ingredients = listOf("apple", "banana")
```
medwardson, line 30:
```suggestion
```
no longer in use

app/src/main/java/com/github/se/polyfit/data/api/RecipeInstructionResponseAPI.kt
```diff
@@ -0,0 +1,90 @@
...
+      val jsonAdapter = moshi.adapter(RecipeInstruction::class.java).lenient()
+      return jsonAdapter.fromJson(jsonString) ?: throw IllegalArgumentException("Invalid JSON data")
+    }
+  }
+}
+
+@JsonClass(generateAdapter = true)
+data class RecipeInstruction(val name: String?, val steps: List<Step>) {
+  companion object {
+    fun faillure(): RecipeInstruction {
```
medwardson, line 28:
```suggestion
    fun failure(): RecipeInstruction {
```

app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt
```diff
@@ -168,4 +170,78 @@ class SpoonacularApiCaller {
...
+        recipe.recipeInformation.instructions = recipeInfo.steps.map { it.step }
+
+        // Add information about the ingredients, removing duplicates
+        val ingredients = recipeInfo.steps.flatMap { it.ingredients ?: emptyList() }
+        recipe.recipeInformation.ingredients = ingredients.distinctBy { it.name }
+      }
+      return recipesResponse.recipes
+    } catch (e: Exception) {
+      Log.e("SpoonacularApiCaller", "Error getting recipe from ingredients", e)
+      return emptyList()
```
medwardson, line 244:
```suggestion
     recipesResponse.recipes
    } catch (e: Exception) {
      Log.e("SpoonacularApiCaller", "Error getting recipe from ingredients", e)
      emptyList()
```

Since you are returning on line 226 you don't need to return here, just have the value

app/src/main/java/com/github/se/polyfit/ui/components/card/RecipeInstructions.kt
```diff
@@ -0,0 +1,36 @@
...
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+@Composable
+fun recipeInstructions(step: String, stepNumber: Int) {
```
medwardson, line 17:
```suggestion
fun RecipeInstructions(step: String, stepNumber: Int) {
```

app/src/main/java/com/github/se/polyfit/ui/screen/recipeRec/RecipeRecommendationMoreDetailScreen.kt
```diff
@@ -41,47 +40,50 @@ fun RecipeRecommendationMoreDetailScreen(
...
+                button2Title = ContextCompat.getString(context, R.string.Recipe),
+                recipeRecViewModel)
+          }
+
+          if (showIngredient) {
+            selectedRecipe?.recipeInformation?.ingredients?.forEach {
+              item { IngredientInfoCard(ingredient = it) }
+            }
+          } else {
+            selectedRecipe!!.recipeInformation.instructions?.forEachIndexed { index, step ->
```
medwardson, line 81:
```suggestion
            selectedRecipe!!.recipeInformation.instructions.forEachIndexed { index, step ->
```
i think this value should always exist, no need for safe assert

app/src/main/java/com/github/se/polyfit/ui/screen/recipeRec/RecipeRecommendationScreen.kt
```diff
@@ -58,15 +62,31 @@ fun recipeDisplay(
...
+@Composable
+fun loader(paddingValues: PaddingValues) {
+
+  Box(
+      modifier = Modifier.fillMaxSize().padding(paddingValues).testTag("Loader"),
+      contentAlignment = Alignment.Center) {
+        CircularProgressIndicator(modifier = Modifier.padding(16.dp).testTag("LoadingPost"))
+      }
+  return
+}
```
medwardson, line 92:
```suggestion
@Composable
fun Loader(paddingValues: PaddingValues) {
  Box(
      modifier = Modifier.fillMaxSize().padding(paddingValues).testTag("Loader"),
      contentAlignment = Alignment.Center) {
        CircularProgressIndicator(modifier = Modifier.padding(16.dp).testTag("LoadingPost"))
      }
}
```
No need to return early here (unless I'm missing something?). Also i think general rule of thumb for composable functions is Title case (all first letter caps). Same for line 43 in this file

medwardson (commented):
Left some syntax/stylistic comments. How should I invoke this screen with the right data to test functionality?

app/src/main/java/com/github/se/polyfit/ui/screen/recipeRec/RecipeRecommendationScreen.kt
```diff
@@ -58,15 +62,31 @@ fun recipeDisplay(
               fontWeight = FontWeight.Bold,
```
medwardson, line 62:
```suggestion
              fontWeight = FontWeight.Bold,
              modifier = Modifier.padding(8.dp)
```
I'd suggest adding some padding between the side of the screen and the text, just to make it visually more appealing. I'd also suggest using another colour like PurpleGrey to make the colour less harsh and more inline with the theming

medwardson (approved):
Functionality wise, things look good! 

Suggestions:
- Include some type of messaging when there's no ingredients/recipe present
- Since bookmarking doesn't exist yet, I'd maybe remove the + button temporarily to avoid confusion

Good job 🥳

app/src/androidTest/java/com/github/se/polyfit/data/api/SpoonacularApiCallerTest.kt
RandomUsername1315: Great suggestion !

app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt
RandomUsername1315: Yep good catch.

app/src/androidTest/java/com/github/se/polyfit/ui/screen/recipeRec/RecipeRecommendationMoreDetailScreenTest.kt
RandomUsername1315: Great recommendation, I will do the same for the UIs

app/src/main/java/com/github/se/polyfit/ui/components/card/RecipeInstructions.kt
Neb1010, line 1: ### Can be improved
- Modifier Chain: The modifier chain can be split for better readability.
- Accessibility: Consider adding content descriptions for accessibility.

Neb1010 (approved):
Good job for your implementation

app/src/androidTest/java/com/github/se/polyfit/ui/screen/recipeRec/MoreDetailScreenScreen.kt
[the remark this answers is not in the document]
RandomUsername1315: The class that performs that actual tests is called that so this class cannot be renamed by the name that you suggested.

app/src/main/java/com/github/se/polyfit/data/api/RecipeInstructionResponseAPI.kt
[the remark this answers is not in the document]
RandomUsername1315: Yeah that's basically it. If I knew about moshi at the beginning of the project, I would have saved a bunch of time.

app/src/main/java/com/github/se/polyfit/data/api/SpoonacularAPI.kt
```diff
@@ -168,4 +165,78 @@ class SpoonacularApiCaller {
       RecipeFromIngredientsResponseAPI.faillure()
     }
   }
+
```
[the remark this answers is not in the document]
RandomUsername1315: Yeah, I 100% agree with you. I should have used `moshi`, but I hadn't discovered it yet, so I implemented the JSON parsing myself. Since I did not modify the previous API call functions, I do not think it is within the scope of this PR to change this.

app/src/main/java/com/github/se/polyfit/ui/components/card/RecipeInstructions.kt
RandomUsername1315: Nice suggestions !

PR #321, opened 2024-05-26
Getting data from sonarcloud

RandomUsername1315:
## Bugfix: CI Reading the Wrong Values

An issue with the CI reading and calculating the line coverage value locally resulted in a lower value being displayed compared to the `SonarCloud` line coverage value. This is probably due to the fact that SonarCloud ignores some files while the local calculation did not. To fix this, the CI now queries the SonarCloud API for the total line coverage and fails the build if it is lower than 80%.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,52 +169,19 @@ jobs:
         # The xmlReportPaths needs to point to where your JaCoCo coverage report is generated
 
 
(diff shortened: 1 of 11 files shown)


antomax33 (approved):
Way better to use this line coverage :+1:

medwardson (approved):
😂 Thank you for finding this

PR #322, opened 2024-05-26
Feature/backend/qrcode/scanning

antomax33:


app/src/main/java/com/github/se/polyfit/viewmodel/qrCode/barcodeViewModel.kt
```diff
@@ -0,0 +1,59 @@
...
+      count = 1
+      return
+    }
+
+    val list = _listId.value?.toMutableList() ?: mutableListOf()
+    list.add(0, id)
+    _listId.value = list
+    Log.v("QrCodeViewModel", "new list: ${_listId.value}")
+  }
+}
```
RandomUsername1315, line 52: Function is a bit hard to understand what it does, I would refactor it this way in order to reduce the cognitive complexity.

```suggestion
fun addId(id: String?) {
    if (id.isNullOrEmpty() || _listId.value!!.contains(id) || id.length !in MIN_BARCODE_LENGTH..MAX_BARCODE_LENGTH) {
        return
    }

    if (id != previousScan) {
        previousScan = id
        count = 1
        return
    }

    if (++count == REQUIRED_SCAN_COUNT) {
        _listId.value = mutableListOf(id).apply { addAll(_listId.value!!) }
        Log.v("QrCodeViewModel", "new list: ${_listId.value}")
    }
}
```

app/src/main/java/com/github/se/polyfit/ui/screen/IngredientsOverview.kt
```diff
@@ -42,16 +47,32 @@ fun IngredientsOverview(
...
       topBar = { SimpleTopBar(title = context.getString(R.string.Product)) { navigateBack() } },
       bottomBar = { BottomBarIngredient(navigateForward = navigateForward) },
       floatingActionButton = { FloatingActionButtonIngredients(onClickFloatingButton) },
       containerColor = MaterialTheme.colorScheme.background,
       modifier = Modifier.testTag("IngredientsOverviewScaffold")) {
-        ListProducts(listIngredients = listProducts, Modifier.fillMaxSize().padding(it))
+        CameraXScreen(barCodeCodeViewModel)
+
+        // TODO THIS IS TEMPORARY, REMOVE IT WHEN ADDING THE REAL VALUES
+        ListProducts(
```
RandomUsername1315, line 72: I do not know if this is in the scope of this PR or maybe a future one, but it would be great if you could remove an ingredient from the list if you add it by accident, similar to how it works in the ingredient screen when creating a meal.

app/src/main/java/com/github/se/polyfit/di/UserModule.kt
```diff
@@ -102,4 +103,10 @@ object UserModule {
   fun providesFirebaseAuth(): FirebaseAuth {
     return FirebaseAuth.getInstance()
   }
+
+  @Provides
+  @Singleton
+  fun providesBarCodeCodeViewModel(): BarCodeCodeViewModel {
+    return BarCodeCodeViewModel()
+  }
```
RandomUsername1315, line 111: A potential issue I noticed is that because you added a `provides` function, only one instance of the ViewModel will be created. This means that if the user scans a few ingredients and proceeds to the recipe recommendation and then goes back to the home screen, or if after scanning they decide to go back, the next time they try to scan ingredients, the previous values will still be there. This happens because the list in the ViewModel is not cleared. This can be easily solved by deleting this `provides` function which will cause a new instance of the ViewModel to be created each time.

PR #324, opened 2024-05-27
Prevent users from having their birthday set in the future

medwardson:
From the feedback we received in Sprint 9 retro, we should block users from setting birthdays in the future.

I added an extra default parameter to the date selector such that if the user passes the non-default value, users can not select a date in the future.


antomax33 (approved):
Lgtm

PR #325, opened 2024-05-27
Fix the squished circle for nutrients

aprilslis:
When making a post, if the screen is too small the circle gets squished

Before:
Screenshot 2024-05-27 at 5 24 12 PM

Results:
PNG image PNG image


medwardson (approved):
Nice and simple, good job

PR #326, opened 2024-05-28
Issue handling weird weight units from Spoonacular

RandomUsername1315:
### Bugfix

There was an issue with how the `MeasurementUnits.kt` class handled non-predefined measurement units and conversion between them. Specifically, there was a problem when trying to convert `CUPS` into `GRAMS`. This was fixed by ensuring that an error is not thrown in this situation and by adding try-catch blocks around the units conversion to prevent the app from crashing. If any issues occur when converting units, a default value of 1.0 is used or an empty list.


app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
```diff
@@ -15,46 +16,66 @@ data class DailyCalorieSummary(val date: LocalDate, val totalCalories: Double)
...
+          val (occasion, _) = pair
+          val value = mealValues.find { it.first == occasion }?.second ?: 0.0
+          Pair(occasion, value)
+        }
       }
+    } catch (e: Exception) {
+      Log.e("LocalDataProcessor", "Error getting calories per meal occasion today", e)
+      MutableLiveData(emptyList())
     }
   }
```
sharif-bbd, line 76: Might be a good idea to create something like 
`val result = MutableLiveData<List<Pair<MealOccasion, Double>>>()`
at the beginning of the function and make a `result.postValue(...)` to set its value inside your try catch block.

sharif-bbd (approved):
Good job fixing this bug !

app/src/main/java/com/github/se/polyfit/data/processor/LocalDataProcessor.kt
RandomUsername1315: That's a great idea! I tried implementing it. Here's what I came up with:

```kotlin
fun getCaloriesPerMealOccasionTodayLiveData(): LiveData<List<Pair<MealOccasion, Double>>> {
    val result = MutableLiveData<List<Pair<MealOccasion, Double>>>()
    
    try {
        val liveData = mealDao.getMealsCreatedOnDateLiveData(LocalDate.now(), user.id)
        liveData.map { meals ->
            val allOccasion =
                MealOccasion.entries.filter { it != MealOccasion.OTHER }.map { Pair(it, 0.0) }

            val mealValues =
                meals
                    .filter { it.isComplete() && it.occasion != MealOccasion.OTHER }
                    .groupBy { it.occasion }
                    .mapValues { (_, meals) -> meals.sumOf { it.calculateTotalCalories() } }
                    .toList()

            allOccasion.map { pair ->
                val (occasion, _) = pair
                val value = mealValues.find { it.first == occasion }?.second ?: 0.0
                Pair(occasion, value)
            }
        }.observeForever { result.postValue(it) }
    } catch (e: Exception) {
        Log.e("LocalDataProcessor", "Error getting calories per meal occasion today", e)
        result.postValue(emptyList())
    }

    return result
}

``` 

The issue with that is the `observeForever` needed in this case, which can potentially cause memory leaks and degrade performance.

PR #330, opened 2024-05-29
Warning when leaving Add Meal

aprilslis:
In this PR, I added a confirmation when leaving `AddMealFlow`. I had to change how we disable the back button on the top bar in this file so that it can be pressed more than once, and also cleaned up the other top bars to be standardized.

Screenshot 2024-05-29 at 2 44 42 PM

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/ui/components/scaffold/SimpleTopBar.kt
+++ b/app/src/main/java/com/github/se/polyfit/ui/components/scaffold/SimpleTopBar.kt
@@ -9,6 +9,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.material3.TopAppBar
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
@@ -23,6 +24,14 @@ import com.github.se.polyfit.ui.utils.titleCase
 @Composable
 fun SimpleTopBar(title: String, navigateBack: () -> Unit) {
   var goBackClicked by remember { mutableStateOf(false) }
+
+  LaunchedEffect(key1 = goBackClicked) {
+    if (goBackClicked) {
+      kotlinx.coroutines.delay(1000)
+      goBackClicked = false
+    }
+  }
+
   TopAppBar(
       title = {
         Text(
@@ -34,10 +43,8 @@ fun SimpleTopBar(title: String, navigateBack: () -> Unit) {
       navigationIcon = {
         IconButton(
             onClick = {
-              if (!goBackClicked) {
-                goBackClicked = true
-                navigateBack()
-              }
+              goBackClicked = true
+              navigateBack()
             },
             content = {
               Icon(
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/se/polyfit/ui/screen/NutritionScreen.kt
```diff
@@ -45,7 +36,7 @@ fun NutritionScreen(
   val isComplete by mealViewModel.isComplete.collectAsState()
 
   Scaffold(
-      topBar = { TopBar(navigateBack = navigateBack) },
+      topBar = { SimpleTopBar(title = "Nutrition Facts", navigateBack = navigateBack) },
```
RandomUsername1315, line 39: Awesome job refactoring this to use the `SimpleTopBar` !

app/src/main/java/com/github/se/polyfit/ui/components/scaffold/SimpleTopBar.kt
```diff
@@ -18,11 +19,25 @@ import androidx.compose.ui.platform.testTag
...
 fun SimpleTopBar(title: String, navigateBack: () -> Unit) {
   var goBackClicked by remember { mutableStateOf(false) }
+
+  // Delay the goBackClicked state change to prevent multiple back navigation.
+  LaunchedEffect(key1 = goBackClicked) {
+    if (goBackClicked) {
+      kotlinx.coroutines.delay(1000)
+      goBackClicked = false
+    }
+  }
```
RandomUsername1315, line 39: Usually it is not a great idea to use a delay in your code. Can't you achieve the same functionality using something like the previous implementation ?

app/src/main/java/com/github/se/polyfit/ui/flow/AddMealFlow.kt
```diff
@@ -43,4 +62,29 @@ fun AddMealFlow(
...
+                onClick = goBack,
+                modifier = Modifier.testTag("GoBack"))
+            PrimaryButton(
+                text = context.getString(R.string.denyRequest),
+                color = PrimaryPink,
+                onClick = closeDialog,
+                modifier = Modifier.testTag("DenyButton"))
+          }
+    }
+  }
```
RandomUsername1315, line 89: As this screen is only relevant to the `IngredientScreen`, it might be a good idea to move it to that file if possible.

RandomUsername1315 (approved):
Awesome job!

app/src/main/java/com/github/se/polyfit/ui/components/scaffold/SimpleTopBar.kt
aprilslis: The previous implementation blocks the button from ever being pressed again, which is not good for what we want. I researched and this is the only suggested way to achieve this. Do you have anything in mind?

aprilslis: I did some research, and it says that delay is bad if it blocks the main thread, since users have to wait longer than needed. But, for a UI element in a non-blocking thread (coroutine) it is ok.

"Yes, if you're using it to wait for a result, that is quite terrible. Would have to see how you're actually using it to understand for sure if there's a problem. delay() in a fragment is totally appropriate if it has to do with UI appearances." from https://stackoverflow.com/questions/76155021/is-it-a-good-practice-to-use-the-kotlin-delay-method-inside-a-fragment

PR #333, opened 2024-05-29
Fix for when location not activated

### LINKED ISSUE #331: Location fix
when we turn off location the app doesn't crash anymore

Neb1010:
If the User turn off the location (not in the setting but on the top drawer), it was leading to a crash, not anymore.


RandomUsername1315 (approved):
Awesome I tested it and the app didn;t crash. Awesome job !

PR #334, opened 2024-05-29
Large username causes the app to crash

RandomUsername1315:
## Bugfix

**Issue:** When the user chooses a large username, it affects the placement of the composable on the overview screen and can sometimes cause the app to crash.

**Resolution:** To fix this, the username is now set to have a maximum of 20 characters.


sharif-bbd (approved):
Good catch ! Nice check to implement on the size of the username!

PR #335, opened 2024-05-29
Bug that caused the values of Height and Weight to be infinity 

RandomUsername1315:
### Bugfix

Fixed an issue where user-entered height and weight values that were too large would be rounded up to infinity. The fix ensures that when Kotlin automatically rounds a number to infinity, the previous valid entry is used instead.

### DIFF
--- a/app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
+++ b/app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
@@ -1,6 +1,5 @@
 package com.github.se.polyfit.ui.screen.settings
 
-import android.util.Log
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
@@ -118,12 +117,7 @@ private fun AccountSettings(
           keyboardOptions = KeyboardOptions.Default.copy(keyboardType = KeyboardType.Number),
           singleLine = true,
           onValueChange = {
-            var update = it.toLongOrNull()
-
-            // If there is a Long overflow, the use the previous value
-            if (update == null && it.toBigDecimalOrNull() != null) {
-              update = height.toLongOrNull()
-            }
+            val update = checkNotInfinity(it, height)
 
             height =
                 when {
@@ -146,15 +140,7 @@ private fun AccountSettings(
           keyboardOptions = KeyboardOptions.Default.copy(keyboardType = KeyboardType.Number),
           singleLine = true,
           onValueChange = {
-            var update = it.toLongOrNull()
-
-            // If there is a Double overflow, the use the previous value
-            Log.d("AccountSettings", "update: $update")
-            Log.d("AccountSettings", "updateBig: ${it.toBigDecimalOrNull()}")
-
-            if (update == null && it.toBigDecimalOrNull() != null) {
-              update = weight.toLongOrNull()
-            }
+            val update = checkNotInfinity(it, weight)
 
             weight =
                 when {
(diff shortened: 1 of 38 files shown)


RandomUsername1315:
@medwardson in theory it should be the exact same fix.

app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
```diff
@@ -117,7 +118,13 @@ private fun AccountSettings(
           keyboardOptions = KeyboardOptions.Default.copy(keyboardType = KeyboardType.Number),
           singleLine = true,
           onValueChange = {
-            val update = it.toLongOrNull()
+            var update = it.toLongOrNull()
+
+            // If there is a Long overflow, the use the previous value
+            if (update == null && it.toBigDecimalOrNull() != null) {
```
antomax33, line 124: Is it useful to check if update is null?
```suggestion
            if (it.toBigDecimalOrNull() != null) {
```

app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
```diff
@@ -162,11 +177,16 @@ private fun AccountSettings(
           keyboardOptions = KeyboardOptions.Default.copy(keyboardType = KeyboardType.Number),
           singleLine = true,
           onValueChange = {
-            val update = it.toLongOrNull()
+            var update = it.toLongOrNull()
+
+            if (update == null && it.toBigDecimalOrNull() != null) {
```
antomax33, line 182: same

app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
```diff
@@ -139,16 +146,24 @@ private fun AccountSettings(
...
           singleLine = true,
           onValueChange = {
-            val update = it.toDoubleOrNull()
+            var update = it.toLongOrNull()
+
+            // If there is a Double overflow, the use the previous value
+            Log.d("AccountSettings", "update: $update")
+            Log.d("AccountSettings", "updateBig: ${it.toBigDecimalOrNull()}")
+
+            if (update == null && it.toBigDecimalOrNull() != null) {
```
antomax33, line 155: same

antomax33 (approved):
Nice to add a test to not have the bug anymore

app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
```diff
@@ -117,7 +118,13 @@ private fun AccountSettings(
           keyboardOptions = KeyboardOptions.Default.copy(keyboardType = KeyboardType.Number),
           singleLine = true,
           onValueChange = {
-            val update = it.toLongOrNull()
+            var update = it.toLongOrNull()
+
+            // If there is a Long overflow, the use the previous value
+            if (update == null && it.toBigDecimalOrNull() != null) {
+              update = height.toLongOrNull()
+            }
```
sharif-bbd, line 126: Since this format is reused throughout this file I would suggest making it a function with a param for the previous value concerned to improve code modularity

sharif-bbd (commented):
Great catch and good job on fixing the bug efficiently

app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
RandomUsername1315: If you try to convert a huge string, for example, "10000000000000000000000000", to a number bigger than the largest possible Long, then the function `.toLongOrNull()` returns null. The issue that was happening is that the null value was interpreted as `infinity`, which looked out of place.

app/src/main/java/com/github/se/polyfit/ui/screen/settings/AccountSettingsScreen.kt
RandomUsername1315: Great suggestion ! Will do !

PR #336, opened 2024-05-30
Make the description for post creation close on enter press

medwardson:
As you can see in the before and after, instead of a new line the enter button is now a checkmark to close the keyboard.
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
      Before Image
    </td>
    <td>
      After Image
    </td>
  </tr>
</table>


RandomUsername1315 (approved):
Awesome job and thanks for adding documentation !

PR #337, opened 2024-05-30
Feature/map/improvement

Neb1010:
#### Front end
- Can directly display a `Post` on the same view by a clic on the `Marker` and to no display any `Post` just clic anywhere on the `GoogleMap` 

image

- `Postview` only shows all `Post` in a radius of 2km (the max the user can)

#### Back end
- Optimization of query on Firebase, if the set of post's keys hasn't changed with a radius change, it will not query a fetch.

### File in comment
[This file](https://github.com/swent-group10/polyfit/pull/337/files#diff-f64f89b650134b3efec5989b32fe083eb46709c18e2d1770a18dff62d79cf2d9) is in comment because if we don't have enough code coverage I will try to fix these tests, as the implementation has changed they are not up to date

### What could be improved for a V2

Initially, the update was done dynamically, but it could happen that the nearest posts would remain displayed even if they were outside the circle. Imagine three layers: layer 1 with 0 posts, layer 2 with 1 post, and layer 3 with 2 posts. If the radius is larger than layer 3, the map will display 3 posts. By reducing the radius, we will be smaller than layer 3, so it will query only the posts from layer 2 and layer 1. However, if we reduce the radius quickly enough to have only layer 1 in the circle, since there are no posts, the query response will be faster than that of layer 3 to 2, resulting in the final query response from 3 to 2.

I tried several ways to solve this small problem, such as using locks or CAS (Compare and Swap). None were very convincing, even though some solutions seemed to reduce the probability of encountering this scenario. I felt more like I was doing code tinkering that depended on the speed of Firebase, and the solution wouldn't scale with the number of posts in the second layer.

So currently, the update is no longer done dynamically to avoid this scenario. I find it a bit less nice, but for a V2, it could be cool to make it dynamic again.

Here a small video of the problem in a dynamically fashion
Vidéo sans titre ‐ Réalisée avec Clipchamp


app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -44,68 +51,83 @@ val MAX_SIZE_RAYON = 2000f
...
+   */
+  val mutex = Mutex()
+  fun onSliderChanged(scope: CoroutineScope) {
+    scope.launch {
+      mutex.withLock {
+        if (mainLaunchedIsFinished) {
+
+          mapViewModel.setRadius(sliderPosition.toDouble() / 1000)
+          mapViewModel.listenToPosts()
+          delay(3000)
```
antomax33, line 114: Set 3000 as an external constant

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -44,68 +51,83 @@ val MAX_SIZE_RAYON = 2000f
...
+   * @param scope The CoroutineScope in which to launch the coroutine.
+   * @param newPosition The new position of the slider.
+   */
+  val mutex = Mutex()
+  fun onSliderChanged(scope: CoroutineScope) {
+    scope.launch {
+      mutex.withLock {
+        if (mainLaunchedIsFinished) {
+
+          mapViewModel.setRadius(sliderPosition.toDouble() / 1000)
```
antomax33, line 112: It is useful to have the units
```suggestion
          mapViewModel.setRadius(sliderPositionMeter.toDouble() / 1000)
```

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -114,39 +136,103 @@ fun MapScreen(
...
-                    contentDescription = "description",
-                    onClick = { goToMarker(goToPost, markerState) })
-              }
+              CircleOnMap(
+                  center = latLng,
+                  radius = doubleSliderPosition + 10,
+                  strokeColor = Color.Black.copy(alpha = 0.2f),
+                  visible = isCircleShown)
+
+              Marker(listMarker) { post -> selectedPost = post }
```
antomax33, line 154: I'm not sure if the order affect who is on the foreground but the marker should be above the circle on the screen

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -114,39 +136,103 @@ fun MapScreen(
...
 
-fun goToMarker(goToPost: () -> Unit, marker: MarkerState): Boolean {
-  goToPost()
-  return true
+@Composable
+@GoogleMapComposable
+fun CircleOnMap(center: LatLng, radius: Double, strokeColor: Color, visible: Boolean) {
+
+  Circle(center = center, radius = radius, strokeColor = strokeColor, visible = visible)
+}
```
antomax33, line 176: Why not directly calling Circle? or importing Circle with alias CircleOnMap

antomax33: Oh I see, you want to have all googlemap drawable at the same place. Why not

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -114,39 +136,103 @@ fun MapScreen(
...
+ * @param listMarker The list of PostMarker objects to be displayed on the map.
+ * @param selectedPost A lambda function that sets the selected post when a marker is clicked.
+ */
+@Composable
+@GoogleMapComposable
+fun Marker(listMarker: List<PostMarker>, selectedPost: (Post) -> Unit) {
+  listMarker.forEach { postMarker ->
+    Marker(
+        state = postMarker.markerState,
+        title = "Post",
```
antomax33, line 194: You can set it in string.xml

antomax33 (commented):
Almost done, well job

app/src/androidTest/java/com/github/se/polyfit/viewmodel/post/ViewPostViewModelTest.kt
```diff
@@ -1,5 +1,6 @@
 package com.github.se.polyfit.viewmodel.post
 
+/*
```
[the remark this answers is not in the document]
Neb1010: explication in preview

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
Neb1010: No, it doesn't affect

app/src/main/java/com/github/se/polyfit/viewmodel/post/ViewPostViewModel.kt
```diff
@@ -19,17 +27,43 @@ import kotlinx.coroutines.flow.StateFlow
...
+    }
+    _location.observeForever { getNearbyPosts(it) }
   }
 
-  fun getAllPost(): LiveData<List<Post>> {
-    return postFirebaseRepository.posts
+  fun getNearbyPosts(location: Location) {
+
+    viewModelScope.launch {
+      withContext(Dispatchers.Main) {
```
RandomUsername1315, line 57: I think that this double dispatcher is catching concurrency issues that sometimes cause the spinner to loop forever. I replaced it with this, and it seems to work:

```suggestion
viewModelScope.launch {
    withContext(Dispatchers.Main) {

File: app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt [hunk 8]
```diff
@@ -24,88 +25,112 @@ import androidx.compose.ui.unit.dp
...
 // Size in meters of the rayon of the circle to see other posts
-val MIN_SIZE_RAYON = 100f
-val MAX_SIZE_RAYON = 2000f
-val EPFL_LATITUDE = 46.5181
-val EPFL_LONGITUDE = 6.5659
+const val MIN_SIZE_RAYON = 100f
+const val MAX_SIZE_RAYON = 2000f
+const val EPFL_LATITUDE = 46.5181
+const val EPFL_LONGITUDE = 6.5659
+const val THREE_SECONDS = 3000L
```
antomax33, line 53: It may be better to call it DELAY_CIRCLE_DISAPPEAR

app/src/main/java/com/github/se/polyfit/viewmodel/post/ViewPostViewModel.kt
```diff
@@ -19,17 +27,43 @@ import kotlinx.coroutines.flow.StateFlow
 @HiltViewModel
 class ViewPostViewModel
 @Inject
-constructor(private val postFirebaseRepository: PostFirebaseRepository) : ViewModel() {
+constructor(
+    private val postFirebaseRepository: PostFirebaseRepository,
+    private val postLocalRepository: PostLocationModel
+) : ViewModel() {
   private val _posts: MutableStateFlow<List<Post>> = MutableStateFlow(mutableListOf())
+  val posts: StateFlow<List<Post>> = _posts
```
RandomUsername1315, line 35: If the other variable were changed to LiveData, it might be a good idea, for the sake of consistency, to also make this one LiveData.

app/src/main/java/com/github/se/polyfit/viewmodel/map/MapViewModel.kt
```diff
@@ -30,19 +34,22 @@ constructor(
   val posts: LiveData<List<Post>> = _posts
 
   init {
-    setRadius(0.0) // default radius can be change later for a better default value
+    setRadius(1.0) // default radius can be change later for a better default value
   }
 
   fun listenToPosts() {
+    Log.d("MapViewModel", "listenToPosts: ")
+    Log.d("MapViewModel", "listenToPosts: ${location.value}")
```
RandomUsername1315, line 42:
```suggestion

```

antomax33 (approved):

app/src/main/java/com/github/se/polyfit/ui/screen/PostInfoScreen.kt
```diff
@@ -61,31 +60,32 @@ fun PostInfoScreenContent(
...
 ) {
-  val posts = viewPostViewModel.getAllPost().observeAsState(initial = listOf())
-  val isFetching by viewPostViewModel.isFetching.collectAsState()
 
   Scaffold(modifier = Modifier.padding(padding)) {
-    if (isFetching) {
+    val isFetching = viewPostViewModel.isFetching.value
+    val posts = viewPostViewModel.posts.collectAsState(initial = emptyList()).value
+    if (isFetching == true) {
+      Log.d("PostInfoScreenContent", "Fetching")
```
RandomUsername1315, line 68:
```suggestion
```

app/src/main/java/com/github/se/polyfit/ui/screen/PostInfoScreen.kt
```diff
@@ -61,31 +60,32 @@ fun PostInfoScreenContent(
     padding: PaddingValues = PaddingValues(),
     viewPostViewModel: ViewPostViewModel = hiltViewModel(),
 ) {
-  val posts = viewPostViewModel.getAllPost().observeAsState(initial = listOf())
-  val isFetching by viewPostViewModel.isFetching.collectAsState()
 
   Scaffold(modifier = Modifier.padding(padding)) {
-    if (isFetching) {
+    val isFetching = viewPostViewModel.isFetching.value
```
RandomUsername1315, line 65: This needs to be collected and observed as a state in order for the UI to be updated dynamically.

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
```diff
@@ -169,6 +174,16 @@ class PostFirebaseRepository(
     val batches = keys.chunked(batchSize)
     var completedBatches = 0
 
+    if (keys.isEmpty()) {
+      completion(posts)
+      return
+    }
+    if (keys == previousNearbyKeys) {
+      Log.d("GeoQuery", "same keys, returning early")
```
RandomUsername1315, line 182:
```suggestion
```

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -24,88 +25,112 @@ import androidx.compose.ui.unit.dp
...
+   */
+  val mutex = Mutex()
+  fun onSliderChanged(scope: CoroutineScope) {
+    scope.launch {
+      mutex.withLock {
+        if (mainLaunchedIsFinished) {
+
+          mapViewModel.setRadius(sliderPositionInMeters.toDouble() / 1000)
+          mapViewModel.listenToPosts()
+          delay(THREE_SECONDS)
```
RandomUsername1315, line 117: In general should not use delays in the code, but this time it is necessary. I would add a comment to say something along the line that it is need to hide the circle radius after a certain time

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -24,88 +25,112 @@ import androidx.compose.ui.unit.dp
...
+   * Effect to be launched when the mapViewModel changes. Initializes the current location and sets
+   * up the map and post listeners.
+   *
+   * This effect fetches the current location asynchronously, sets the location in the mapViewModel,
+   * and begins listening to posts. It also updates the camera position of the map and observes
+   * posts to update the state.
+   *
+   * @param mapViewModel The ViewModel that handles map-related data and operations.
+   */
+  LaunchedEffect(key1 = mapViewModel) {
```
RandomUsername1315, line 87: So, because every time the user navigates out of the map view and comes back, a new model will be injected, instead of using the mapViewModel, you could use a `Unit` to make the code a bit easier to understand:

```suggestion
LaunchedEffect(Unit) {

```

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -1,21 +1,22 @@
 package com.github.se.polyfit.ui.screen
 
 // import the colors PrimaryPink
-import android.util.Log
+
```
RandomUsername1315, line 4: I know that you wanted to group everything related to the map in a single file, but it feels a bit too clustered. I would recommend breaking it down into smaller components.

RandomUsername1315 (commented):
Awesome job and thanks for adding documentation.

app/src/main/java/com/github/se/polyfit/viewmodel/post/ViewPostViewModel.kt
Neb1010: I tried it doesn't work

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
```diff
@@ -159,6 +161,9 @@ class PostFirebaseRepository(
    * @param postCollection The Firestore collection to fetch the posts from.
    * @param completion The callback function to call once the posts have been fetched.
    */
+  private var previousNearbyKeys: List<String> = emptyList()
+  private var previousListOfPosts: List<Post> = emptyList()
+
```
sharif-bbd, line 166: It seems to me that the value of `previousNearbyKeys` never changes throughout the function so I don't understand why we have an if statement `if(keys == preivousNearbyKeys)`

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
Neb1010: how do you make a delay without using delay?

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
Neb1010: I tried both, it doesn't change at all

app/src/androidTest/java/com/github/se/polyfit/viewmodel/post/ViewPostViewModelTest.kt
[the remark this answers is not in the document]
Neb1010: see preview

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
```diff
@@ -187,6 +202,7 @@ class PostFirebaseRepository(
 
                 completedBatches++
                 if (completedBatches == batches.size) {
+                  previousListOfPosts = posts
                   completion(posts)
                 }
```
sharif-bbd, line 207: I have read in a few sources such as these :
https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#mutual-exclusion
https://jacquessmuts.github.io/post/coroutine_sync_mutex/

That within a coroutine scope the use of synchronization is better done with a `MutexLock` paired with `withContext()` as issues could arise in the Coroutine, however considering this is a V1 and the code currently doesn't seem to be having any issues I think it should be fine to let it as it is. Just suggesting for later uses

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -24,88 +25,112 @@ import androidx.compose.ui.unit.dp
...
-    goToPost: () -> Unit,
-    mapviewMap: MapViewModel = hiltViewModel()
-) {
+fun MapScreen(paddingValues: PaddingValues, mapViewModel: MapViewModel = hiltViewModel()) {
 
   val posts = remember { mutableStateOf(listOf<Post>()) }
+
   val currentLocation = remember { mutableStateOf(Location.default()) }
-  val mutex = Mutex()
+
```
sharif-bbd, line 62: It might be a good initiative to use viewModel states to declare these values and collect them as state in the UI but it isn't that important at the moment, could be modified at a later notice

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
```diff
@@ -114,39 +139,102 @@ fun MapScreen(
...
 
-fun goToMarker(goToPost: () -> Unit, marker: MarkerState): Boolean {
-  goToPost()
-  return true
+@Composable
+@GoogleMapComposable
+fun CircleOnMap(center: LatLng, radius: Double, strokeColor: Color, visible: Boolean) {
+
+  Circle(center = center, radius = radius, strokeColor = strokeColor, visible = visible)
+}
```
sharif-bbd, line 179: This is a great initiative!

could add optional parameters for more versatility
```
@Composable
@GoogleMapComposable
fun CircleOnMap(center: LatLng, radius: Double, strokeColor: Color, visible: Boolean, alpha: Float = 1f) {
    Circle(center = center, radius = radius, strokeColor = strokeColor.copy(alpha = alpha), visible = visible)
}
```
But since they aren't used currently it doesn't really change much

app/src/main/java/com/github/se/polyfit/ui/screen/PostInfoScreen.kt
```diff
@@ -61,31 +60,32 @@ fun PostInfoScreenContent(
...
+    val posts = viewPostViewModel.posts.collectAsState(initial = emptyList()).value
```
sharif-bbd, line 66: This works well and makes usage of the viewModel, but a suggestion I would have is to declare it as
`val posts by viewPostViewModel.posts.collectAsState(emptyList())`

sharif-bbd (commented):
Great Improvement to the Map feature ! Good job!

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
Neb1010: this is the code of a previous PR

app/src/main/java/com/github/se/polyfit/ui/screen/MapScreen.kt
Neb1010: ?

app/src/main/java/com/github/se/polyfit/data/remote/firebase/PostFirebaseRepository.kt
Neb1010: I'm sorry you unfurtunatly review a outdated code for this part

PR #340, opened 2024-05-30
fix cursor bug, return if no location

antomax33:


app/src/main/java/com/github/se/polyfit/viewmodel/map/MapViewModel.kt
```diff
@@ -35,7 +35,7 @@ constructor(
 
   fun listenToPosts() {
 
-    assert(_location.value != null)
+    if (_location.value == null) return
```
sharif-bbd, line 38: Great Catch on this mistake !

sharif-bbd (approved):

PR #341, opened 2024-05-30
Adjust the MealTag onclick

medwardson:
Our mealtag onclick was empty on the daily recap, making the button clickable with no effect. This overrode the outer onclick for the whole meal to go and edit it. This fix addresses that issue, removing the onclick from the meal tag if it is display only.


RandomUsername1315 (approved):
Nice Job catching that bug !

PR #342, opened 2024-05-30
Limit tag length to 50 chars

medwardson:
To prevent users dumping data into our DB, limit the input on meal tags to 50 chars


RandomUsername1315 (approved):
Nice and short ! Awesome!

PR #344, opened 2024-05-30
Replace search bar with text field

aprilslis:
Originally, we implemented our add ingredient with a search bar to query for existing ingredients. This feature never was developed, so, to clear any confusion with the UI let's use a text field instead.


medwardson (approved):
This lgtm, thx for adding the documentation for the file too

antomax33 (approved):
LGTM

PR #356, opened 2024-06-01
if food not founded, we can retry scanning it

antomax33:
If no food is founded or an network error, the user can scan again the ingredient.

PS: Adding a vibration when detecting a barcode


app/src/test/java/com/github/se/polyfit/viewmodel/qrCode/BarCodeCodeViewModelTest.kt
```diff
@@ -48,7 +48,13 @@ class BarCodeCodeViewModelTest {
     for (i in 1..REQUIRED_SCAN_COUNT) {
       viewModel.addId("123456")
     }
-    assertEquals(listOf("123456"), viewModel.listId.value)
+
+    var i = 0
+    while (listOf("123456") != viewModel.listId.value) {
+      i++
+      sleep(100)
+      assert(i < 100)
```
RandomUsername1315, line 56: Maybe a not a good idea to have a sleep. I think that we lost points during the last milestones do to that.

app/src/main/java/com/github/se/polyfit/ui/screen/CameraXScreen.kt
```diff
@@ -102,3 +108,10 @@ fun CameraXScreen(
     preview.setSurfaceProvider(previewView.surfaceProvider)
   }
 }
+
+fun VibratePhone(context: Context) {
+  val vibrator = context.getSystemService(Context.VIBRATOR_SERVICE) as Vibrator
+
+  val effect = VibrationEffect.createOneShot(100, VibrationEffect.DEFAULT_AMPLITUDE)
+  vibrator.vibrate(effect)
```
RandomUsername1315, line 116: So cool !

RandomUsername1315 (approved):
Such a cool feature to add

